PR #242, opened 2015-01-06
Support for Foo::SomeMessage.decode_from(stream)

bdferris:
The Serialization documentation:
  https://github.com/localshred/protobuf/wiki/Serialization
implies that the following idioms should work for decoding an object:
  Foo::User.decode(bytes)
  Foo::User.decode_from(stream)
However, in practice, decode_from(...) only seems to be supported as an instance
method, not a class method.  By comparison, decode(...) is supported as both an
instance method and a class method.  Let's fix that.

### DIFF
--- a/lib/protobuf/message/serialization.rb
+++ b/lib/protobuf/message/serialization.rb
@@ -11,6 +11,10 @@ module Protobuf
           new.decode(bytes)
         end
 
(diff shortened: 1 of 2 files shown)


localshred:
Nice, good catch. :+1:

PR #244, opened 2015-01-16
Abrandoned/broker probs

abrandoned:
move check available rcv/snd timeout on zeromq to use the ENV if specified (instead of if specified and greater than 200ms, that's a long time in socket connection)

In running the zeromq client/server the most common "failure" is when the broker either doesn't start or fails after running and the servers freezes. This PR adds a check in the running loop of the server to restart the broker if it fails, a new context is required to restart a failed broker.

I also removed the "ClassLength" check in rubocop as we are trying to get the zeromq stuff out of the mainline gem (which we can refactor at that time) and creating an abstraction to remove class length seems prone to creating the wrong abstractions if it is forced.

@localshred @liveh2o

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,9 @@ Lint/EndAlignment:
 Lint/Loop:
   Enabled: false
 
+Metrics/ClassLength:
+  Enabled: false
+
 Style/CaseIndentation:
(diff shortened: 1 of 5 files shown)


liveh2o:
Looks like a good enough cut. :+1: for refactoring this once it's extracted.

:shipit:

PR #245, opened 2015-01-27
CI improvements

tamird:
@localshred @abrandoned @liveh2o

### DIFF
--- a/install-protobuf.sh
+++ b/install-protobuf.sh
@@ -2,9 +2,27 @@
 
 set -ex
 
-version=2.6.1
-basename=protobuf-$version
+gdie() {
+    echo "$@" >&2
+    exit 1
+}
 
-curl -sL https://github.com/google/protobuf/releases/download/v$version/$basename.tar.bz2 | tar jx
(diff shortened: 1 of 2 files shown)


install-protobuf.sh
```diff
@@ -1,8 +1,10 @@
+#!/usr/bin/env sh
+
 set -ex
 
-basename=protobuf-2.6.0
-tarball=$basename.tar.bz2
+version=2.6.1
```
localshred: You could easily take an arg here and default to 2.6.1:

```bash
version=${1:-"2.6.1"}
```

Thus you could install whichever version you wish without having to update the script.

tamird: we'd still have to update it to move the default along to keep up with latest. not sure i see the value

tamird:
ping! happy to close this if y'all aren't interested

tamird:
@localshred this adds testing with protobuf 3 (currently fails, but still)

tamird:
@localshred shall I close this?

PR #250, opened 2015-02-26
Allow the Message to be constructed with a block.

filterfish:
We commonly use the following idiom:

```ruby
Protobuf::Message.new.tap do |p|
  p.field = x
  ...
end
```

This patch allows the following:
```ruby
Protobuf::Message.new do |p|
  p.field = x
  ...
end
```

### DIFF
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -26,11 +26,15 @@ module Protobuf
     # Constructor
     #
 
-    def initialize(fields = {})
+    def initialize(fields = {}, &blk)
       @values = {}
 
-      fields.to_hash.each_pair do |name, value|
-        self[name] = value
+      if blk
+        blk.call(self)
+      else
+        fields.to_hash.each_pair do |name, value|
+          self[name] = value
+        end
       end
     end
 
--- a/spec/lib/protobuf/message_spec.rb
+++ b/spec/lib/protobuf/message_spec.rb
@@ -175,6 +175,12 @@ RSpec.describe Protobuf::Message do
       expect(test_enum.non_default_enum).to eq(2)
     end
 
+    it "initializes with an object with a block" do
+      hashie_object = OpenStruct.new(:to_hash => { :non_default_enum => 2 })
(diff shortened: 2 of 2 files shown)


lib/protobuf/message.rb
```diff
@@ -26,11 +26,15 @@ def self.to_json
     # Constructor
     #
 
-    def initialize(fields = {})
+    def initialize(fields = {}, &blk)
```
localshred: Rename `blk` to `block`.

lib/protobuf/message.rb
```diff
@@ -26,11 +26,15 @@ def self.to_json
...
       @values = {}
 
-      fields.to_hash.each_pair do |name, value|
-        self[name] = value
+      if blk
+        blk.call(self)
+      else
+        fields.to_hash.each_pair do |name, value|
```
localshred: I would prefer that we always utilize the given fields hash (if any given), in addition to calling the block. Perhaps:

```ruby
def initialize(fields = {}, &block)
  fields.to_hash.each_pair do |name, value|
    self[name] = value
  end
  block.call(self) if block_given?
end
```

Which would enable you to do this:

```ruby
foo = Foo.new(:bar => 1) do |f|
  f.baz = 2
end
```

localshred:
Other than those changes, looks ok to me. :+1:

filterfish:
Good one.

I did think about doing it as you've done but I thought my version would be slightly faster. However I've just benchmarked both and the there is no discernible difference!

lib/protobuf/message.rb
```diff
@@ -26,11 +26,15 @@ def self.to_json
...
     #
 
-    def initialize(fields = {})
+    def initialize(fields = {}, &blk)
       @values = {}
 
-      fields.to_hash.each_pair do |name, value|
-        self[name] = value
+      if blk
+        blk.call(self)
```
tamird:
```ruby
if block_given?
  yield self
else
```
is much more performant

Issue #251, opened 2015-02-27
downcasting of values during field assignment


skippy:
hello,

I'm noticing that if I assign an object that inherits from the ruby class which is created from the protocol definition, it is downcast, and I don't see any particular reason why.  For example

Lets say I have the following classes.  (The reason I'm doing this is I'm injecting a *ton* of validations into the new class; the other way to do this is to not have `Corp::Location` inherit from the proto-based ruby object but instead just serialize into it.... such as `Corp::Location#to_proto` calls `Corp::WireDefinition::Location.new(self.to_hash).to_proto`... thoughts on approach?)

Anyway, I digress a bit... check this out:

```ruby
module Corp
  class Location < Corp::WireDefinition::Location
    #a ton of validations
  end
end

module Corp
  class Business < Corp::WireDefinition::Business
    # includes an array of Locations
  end
end
```

if I run the following, on assignment, the `Corp::Location` instance is downcast.  This is problematic if I have methods within `Corp::Location` that I need to call later (such as running validations).

```ruby
business = Corp::Business.new
business.locations = [Corp.Location.new, Corp.Location.new]
business.locations.first.is_a?(Corp.Location) # => false; it is a Corp::WireDefinition::Location
```

The culprit is this line: https://github.com/localshred/protobuf/blob/master/lib/protobuf/field/field_array.rb#L70

I don't understand why it is creating a new object here, downcasting it in the process, and git blame doesn't shed any light as to why this logic is here.  Thoughts?  Would it be ok if I submit a PR that rips this out?

PR #252, opened 2015-03-04
don't downcast when assigning to a field_array

skippy:


skippy:
addresses #251

lib/protobuf/field/message_field.rb
```diff
@@ -9,7 +9,7 @@ class MessageField < BaseField
       #
 
       def acceptable?(val)
-        unless val.instance_of?(type_class) || val.respond_to?(:to_hash)
+        unless val.is_a?(type_class) || val.is_a?(Hash)
```
skippy: this may be a bit outside the scope of this fix.  The reason I made this change was because this [line](https://github.com/localshred/protobuf/pull/252/files#diff-1de145c39faa7b92fc0a33ea3d9fc49bR69) was failing without it.  IT seemed wrong that an incorrect object that has no relationship to the expected object should be acceptable.  It was passing this method because of `val.respond_to?(:to_hash)`, which seems too open ended.  Also, using `instance_of?` doesn't take into account class inheritance while `is_a?` does.

skippy:
just a polite ping :)  Let me know if you want or need additional tests or documentation.  thanks!

lib/protobuf/field/message_field.rb
localshred: I can see the change for `is_a?` but the swap from `respond_to` seems unnecessary. We use duck type checks all over in this lib purposefully. You should be able to pass a hash-like object here. If you change that back I'll get this merged.

lib/protobuf/field/field_array.rb
```diff
@@ -66,6 +66,8 @@ def normalize(value)
 
         if field.is_a?(::Protobuf::Field::EnumField)
           field.type_class.fetch(value)
+        elsif field.is_a?(::Protobuf::Field::MessageField) && value.is_a?(field.type_class)
```
localshred: Ahh, makes sense. Good spot.

localshred:
Sorry for the radio silence, if you can revert that respond_to change I'll get this merged.

lib/protobuf/field/message_field.rb
skippy: hmmm, if I move this to `unless val.is_a?(type_class) || val.respond_to?(:to_hash)`, then  this [test](https://github.com/localshred/protobuf/pull/252/files#diff-1de145c39faa7b92fc0a33ea3d9fc49bR69) fails.

I put that test in there because assigning an unrelated object to a MessageField and having it pass (and possibly drop fields silently if they don't match) seems wrong.  But if that is ok or the desired behavior I'll go ahead and make the change and remove the test.

skippy:
@localshred no worries!  I understand you have a life outside of your OS projects :)

lib/protobuf/field/message_field.rb
localshred: The only reason I think that's ok is because you could do `OtherBasicMessage.new.to_hash` and pass it down... the message will ignore any fields it doesn't know about so it seems ok to me. Thoughts?

skippy:
@localshred and my apologies for going silent!  If the changes to `#to_hash` causes issues, let me know

Issue #249, opened 2015-02-24
Inconsistent field serialization


liveh2o:
I'm seeing some inconsistencies in how we handle field serialization for different field types, which seems wrong.

Given the following definition:

```protobuf
message Account {
  optional double balance = 1
  optional int64 balance_updated_at = 2  
  optional int64 transactions_count = 3
}
```

Protobuf gladly accepts a string for double fields, presumably because strings respond to `to_f`:

```
2.2.0 > a = Account.new
=> #<Account balance=0.0 transactions_count=0 balance_updated_at=0
2.2.0 > a.balance = "123"
=> 123.0 
```

However, it does not accept a string for integer fields, even though strings respond to `to_i`:

```
2.2.0 > a = Account.new
=> #<Account balance=0.0 transactions_count=0 balance_updated_at=0
2.2.0 > a.transactions_count = "1"
TypeError: Unacceptable value 1 for field transactions_count of type Protobuf::Field::Int64Field
```

Curiously, it will accept a time for integer fields, but does not convert it to an integer, even though it does so with strings on double fields:

```
2.2.0 > a = Account.new
=> #<Account balance=0.0 transactions_count=0 balance_updated_at=0
2.2.0 > a.balance_updated_at = Time.current
=> Tue, 24 Feb 2015 20:09:08 UTC +00:00
```

This, of course, becomes a problem once it attempts to encode the message:

```
2.2.0 > a = Account.new(:balance_updated_at => Time.current)
=> #<Account balance=0.0 transactions_count=0 balance_updated_at=Tue, 24 Feb 2015 20:09:08 UTC +00:00
2.2.0 > a.encode
NoMethodError: undefined method `&' for Tue, 24 Feb 2015 21:17:58 UTC +00:00:Time
  ...
```

There may actually be two issues here, but I wanted to document the inconsistent behavior so we can start discussion a possible solution. If we need to split this into two, I'm happy to do so.

localshred:
Wow, that is shocking. Definitely should get some test cases in place for these.

PR #253, opened 2015-03-30
Adds support to specify integer and bool proto fields as string.

hrsht:
Also fixes the range check for integer types.

r: @zachmargolis

### DIFF
--- a/lib/protobuf/field/bool_field.rb
+++ b/lib/protobuf/field/bool_field.rb
@@ -17,7 +17,17 @@ module Protobuf
       # #
 
       def acceptable?(val)
-        [true, false].include?(val)
+        [true, false].include?(val) || %w(true false).include?(val)
+      end
+
+      def coerce!(val)
+        if val == 'true'
+          true
+        elsif val == 'false'
+          false
+        else
+          val
+        end
       end
 
       def decode(value)
--- a/lib/protobuf/field/varint_field.rb
+++ b/lib/protobuf/field/varint_field.rb
@@ -37,9 +37,15 @@ module Protobuf
       #
 
       def acceptable?(val)
-        (val > self.class.min || val < self.class.max)
+        int_val = coerce!(val)
+        int_val >= self.class.min && int_val <= self.class.max
       rescue
         false
+      end
+
+      def coerce!(val)
+        return val.to_i if val.is_a?(Numeric)
+        Integer(val, 10)
       end
 
       def decode(value)
--- a/spec/lib/protobuf/field/bool_field_spec.rb
+++ b/spec/lib/protobuf/field/bool_field_spec.rb
@@ -1,0 +1,51 @@
+require 'spec_helper'
+
+RSpec.describe Protobuf::Field::Int32Field do
+
+  class SomeBoolMessage < ::Protobuf::Message
+    optional :bool, :some_bool, 1
+  end
+
+  let(:instance) { SomeBoolMessage.new }
+
+  describe '#define_setter' do
+    subject { instance.some_bool = value; instance.some_bool }
+
+    [true, false].each do |val|
+      context "when set with #{val}" do
+        let(:value) { val }
+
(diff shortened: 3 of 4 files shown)


lib/protobuf/field/bool_field.rb
```diff
@@ -17,7 +17,15 @@ def self.default
       # #
 
       def acceptable?(val)
-        [true, false].include?(val)
+        [true, false].include?(val) || (val.is_a?(String) && %w(true false).include?(val))
```
zachmargolis: IMO the class checking is not necessary, you could remove that and everything else would work

lib/protobuf/field/bool_field.rb
```diff
@@ -17,7 +17,15 @@ def self.default
...
+      end
+
+      def coerce!(val)
+        if val.is_a?(String)
```
zachmargolis: I tend to dislike checking for particular classes since that defeats the purpose of duck typing. Could do something as simple as `val.to_s == 'true'` if you wanted?

lib/protobuf/field/varint_field.rb
```diff
@@ -37,11 +37,17 @@ def self.encode(value)
       #
 
       def acceptable?(val)
-        (val > self.class.min || val < self.class.max)
+        int_val = coerce!(val)
+        !(int_val < self.class.min || int_val > self.class.max)
```
zachmargolis: do you think `&&` might read more clearly?

```
(self.class.min < int_val && int_val < self.class.max)
```

zachmargolis:
This is super helpful, thanks!

zachmargolis:
Also looks like this fixes #249

lib/protobuf/field/bool_field.rb
hrsht: Good point. Will remove.

lib/protobuf/field/bool_field.rb
hrsht: Will do.

lib/protobuf/field/varint_field.rb
hrsht: Yup, will change. I thought of it myself after sending the PR.

hrsht:
Addressed comments. PTAL?

cc: @localshred

localshred:
Can you provide a reason why this use case should exist? I'm not convinced we should support magically coercing strings into bools or ints.

zachmargolis:
@localshred well one reason would be to fix the inconsistent serialization issue (#249) :grin: 

Another would be that we often build protobuf messages out of query params, where everything is stringified, and moving this coercion here matches our expectations as customers of this gem.

lib/protobuf/field/varint_field.rb
```diff
@@ -37,11 +37,17 @@ def self.encode(value)
...
       def acceptable?(val)
-        (val > self.class.min || val < self.class.max)
+        int_val = coerce!(val)
+        int_val >= self.class.min && int_val <= self.class.max
       rescue
         false
       end
 
+      def coerce!(val)
+        return val.to_i if val.is_a?(Numeric)
```
abrandoned: Is this required? (the .to_i and is_a?(Numeric) check)

we have been looking at standardizing serialization with this branch
https://github.com/localshred/protobuf/compare/bdewitt/fix_coercion

hrsht: It is required to convert a float to an integer, unless that should be disallowed.

Also, `Integer('0x10')` produces 16. IMO, it should return an error and hence requires `Integer(val, 10)` which throws an error if `val` is of type Numeric.

hrsht:
Ping!

hrsht:
Pinging again!

Please let me know if there are any additional changes to address.

localshred:
Sorry @hrsht, was out of town all last week. Will look into this some more this week.

PR #254, opened 2015-04-07
restart service listener if it dies

abrandoned:
the service directory thread can die (for a myriad of reasons), this PR restarts when a lookup request is made if the service directory has been stopped


abrandoned:
@localshred let me know and I can pull this in with the other coercion PR and release a new version if you want

PR #260, opened 2015-07-01
Fix CLI Options bug and RPC Services bug

film42:
Thor's `HashWithIndifferentAccess` doesn't implement `symbolize_keys`, and recently became an issue because the keys would remain strings, instead of being converted into symbols. So, this makes sure we call `to_hash` before we call `symbolize_keys`. To be safe, I added the same `to_hash.symbolize_keys` to the constructors of `ZmqRunner` as well as `SocketRunner`. I also updated the functional zmq_server_spec to make sure everything is :ok_hand: .

Another small bug was our usage of `subclasses` in `Rpc::Service`. When loading rpc_server with rails, everything works fine, because the entire active_support gem is loaded. But if you load an environment file that doesn't load active_support, you'll get a method not found error. To fix this, I explicitly require `active_support/core_ext/class` and everything is :ok_hand: .

One last thing. The link to travis in the README is broken. This corrects that too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # protobuf
 
 [![Gem Version](https://badge.fury.io/rb/protobuf.svg)](http://badge.fury.io/rb/protobuf)
-[![Build Status](https://secure.travis-ci.org/localshred/protobuf.svg?branch=master)](https://travis-ci.org/localshred/protobuf)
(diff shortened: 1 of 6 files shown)


film42:
PS: Specs are passing, :cop: is complaining about problems in master.

PR #261, opened 2015-07-01
a newer rubocop was released and this updates the code to stay inline…

abrandoned:
… with new cops/updates

fixes the rubocop errors that @film42 reported

### DIFF
--- a/lib/protobuf/message/fields.rb
+++ b/lib/protobuf/message/fields.rb
@@ -12,8 +12,10 @@ module Protobuf
           :get_field_by_tag => :get_field,
         )
 
-        def inherited(subclass)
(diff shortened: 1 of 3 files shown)


film42:
:shipit:

lib/protobuf/message/fields.rb
```diff
@@ -12,8 +12,10 @@ def self.extended(other)
           :get_field_by_tag => :get_field,
         )
 
-        def inherited(subclass)
-          inherit_fields!(subclass)
+        other.class_eval do
```
brianstien: This is simply defining a class method on the extending class.  It can be moved to the ClassMethods like the other class methods we are defining on the extending class.

PR #262, opened 2015-07-02
Thread can be sleeping under MRI, fix by using `alive?`

film42:
For some reason, the broker thread will be sleeping under MRI and therefore will fail the `!@broker_thread.stop?` condition. Switching to `@broker_thread.alive?` fixes this and allows protobuf to function normally.


liveh2o:
Any idea why Travis is failing?

film42:
@abrandoned is fixing them in #261

PR #264, opened 2015-09-25
Optimized valid_tag?, enums_for_tag and enums_for_tags

zanker:
We have some rather large protos that make heavy use of enums, and we were seeing fairly slow deserialization times. I ran it through a ruby-prof trace to see if there was anything that could be done to speed up deserialization.
We have some RPC responses that can return a lot of data and use enums. We're seeing fairly slow (around 1,000ms) deserialization times so I ran a few decodes through ruby-prof to see if we could get any quick wins.

Here's what the trace looked like (Ruby 2.2.3) before, I'm only showing things that took 1%+ of the time.

```
Measure Mode: wall_time
Thread ID: 70093065552380
Fiber ID: 70093084071700
Total: 4.599523
Sort by: self_time

 %self      total      self      wait     child     calls  name
 16.25      0.748     0.748     0.000     0.000  2031000   Protobuf::Enum#to_i
 15.35      1.454     0.706     0.000     0.748    30000   Array#select
  7.30      0.745     0.336     0.000     0.410   140000   <Class::Protobuf::Decoder>#read_varint
  3.44      0.176     0.158     0.000     0.018    94000   Kernel#respond_to?
  3.32      1.156     0.152     0.000     1.003    72000   <Class::Protobuf::Decoder>#read_field
  2.44      0.508     0.112     0.000     0.396    72000   <Class::Protobuf::Decoder>#read_key
  2.26      0.298     0.104     0.000     0.194    72000   Protobuf::Message::Fields::ClassMethods#get_field
  2.00      0.159     0.092     0.000     0.067   166500   IO::generic_readable#readbyte
  1.76      0.124     0.081     0.000     0.043   171500   Numeric#nonzero?
  1.66      0.076     0.076     0.000     0.000   410500   Fixnum#&
  1.58      0.073     0.073     0.000     0.000   144000   Protobuf::Field::BaseField#repeated?
  1.48      0.068     0.068     0.000     0.000    69000   Protobuf::Field::BaseField#setter
  1.46      0.067     0.067     0.000     0.000   166500   StringIO#getbyte
  1.19      0.055     0.055     0.000     0.000     1500   Kernel#caller
  1.17      0.054     0.054     0.000     0.000    72000   Protobuf::Message::Fields::ClassMethods#field_store
  1.05      0.087     0.048     0.000     0.039    72000   Protobuf::Field::BaseField#packed?
  1.00      0.137     0.046     0.000     0.091    36003   Class#new
```

With the PR, it now looks like:

```
Measure Mode: wall_time
Thread ID: 70271059223040
Fiber ID: 70271070182440
Total: 3.308983
Sort by: self_time

 %self      total      self      wait     child     calls  name
 10.16      0.736     0.336     0.000     0.400   140000   <Class::Protobuf::Decoder>#read_varint
  4.92      0.182     0.163     0.000     0.019    94000   Kernel#respond_to?
  4.62      1.148     0.153     0.000     0.995    72000   <Class::Protobuf::Decoder>#read_field
  3.33      0.501     0.110     0.000     0.391    72000   <Class::Protobuf::Decoder>#read_key
  3.25      0.308     0.108     0.000     0.201    72000   Protobuf::Message::Fields::ClassMethods#get_field
  2.70      0.156     0.089     0.000     0.066   166500   IO::generic_readable#readbyte
  2.37      0.078     0.078     0.000     0.000   410500   Fixnum#&
  2.34      0.119     0.077     0.000     0.042   171500   Numeric#nonzero?
  2.29      0.076     0.076     0.000     0.000   144000   Protobuf::Field::BaseField#repeated?
  2.12      0.070     0.070     0.000     0.000    69000   Protobuf::Field::BaseField#setter
  2.05      0.068     0.068     0.000     0.000     1500   Kernel#caller
  2.00      0.066     0.066     0.000     0.000   166500   StringIO#getbyte
  1.69      0.056     0.056     0.000     0.000    72000   Protobuf::Message::Fields::ClassMethods#field_store
  1.53      0.093     0.050     0.000     0.042    72000   Protobuf::Field::BaseField#packed?
  1.41      0.102     0.047     0.000     0.055    51000   Kernel#dup
  1.37      0.045     0.045     0.000     0.000    47000   StringIO#read
  1.34      0.138     0.044     0.000     0.093    36003   Class#new
  1.34      0.285     0.044     0.000     0.241    43000   <Class::Protobuf::Decoder>#read_length_delimited
  1.34      0.165     0.044     0.000     0.121    30000   <Class::Protobuf::Enum>#fetch
  1.26      0.042     0.042     0.000     0.000   172000   Fixnum#zero?
  1.16      0.099     0.039     0.000     0.061    25000   Protobuf::Field::StringField#decode
  1.11      0.061     0.037     0.000     0.025    17501   Protobuf::Message#initialize
  1.06      0.035     0.035     0.000     0.000    89501   StringIO#eof?
  1.03      0.034     0.034     0.000     0.000   166500   Fixnum#<<
```

Benchmarking locally, we saw around a 23% improvement to performance on this.


lib/protobuf/enum.rb
```diff
@@ -120,7 +127,8 @@ def self.enum_for_name(name)
     #   Enums, the first enum defined will be returned.
     #
     def self.enum_for_tag(tag)
-      enums_for_tag(tag).first
+      value = mapped_enums[tag.to_i]
```
zachmargolis: `nil.to_i` is zero ... I don't think we want that here. can we guard on `if tag` ?

lib/protobuf/enum.rb
```diff
@@ -198,7 +206,7 @@ def self.name_for_tag(tag)
     # Returns a boolean.
     #
     def self.valid_tag?(tag)
-      tag.respond_to?(:to_i) && all_tags.include?(tag.to_i)
+      tag.respond_to?(:to_i) && mapped_enums.key?(tag.to_i)
```
xaviershay: is `respond_to?` usually expected to be `true`? If yes, maybe catching `NoMethodError` is preferable - should be faster in typical case, slower in error case.

lib/protobuf/enum.rb
```diff
@@ -120,7 +127,8 @@ def self.enum_for_name(name)
     #   Enums, the first enum defined will be returned.
     #
     def self.enum_for_tag(tag)
-      enums_for_tag(tag).first
+      value = mapped_enums[tag.to_i]
+      value ? value.first : nil
```
xaviershay: is it an error to have more than one enum value? Should this pre-emptively raise on that case instead? (likely out of scope for this PR)

zachmargolis: No because the spec allows aliasing enums:

> You can define aliases by assigning the same value to different enum constants. To do this you need to set the allow_alias option to true, otherwise protocol compiler will generate an error message when aliases are found.

source: https://developers.google.com/protocol-buffers/docs/proto?hl=en#enum

lib/protobuf/enum.rb
zanker: Nice catch, if you look at the diff for `self.enums_for_tag`, it was calling `tag.to_i` already. Looks like this might be a pre-existing bug?

zanker:
Looks like the test failures are related to Rubocop primarily, not quite sure how what I did is triggering those failures are it's totally unrelated files?

lib/protobuf/enum.rb
zanker: It looks like `valid_tag?` is a public API that's not used internally, so it wouldn't speed up performance in (our) case. It's probably true that a lot of cases could be replaced with rescuing `NoMethodError` if it's rare that invalid data is passed. Would leave that up to the owners if they're interested before I go dig through that.

lib/protobuf/enum.rb
zachmargolis: eek, thats unfortunate

mmmries:
@zanker if the rubocop errors are on files that haven't been changed it is probably because rubocop has added new checks that we weren't checking against in the past. We could disable those checks in the `.rubocop.yml` file if needed.

zanker:
@mmmries I added it to the TODO so those will go green again

lib/protobuf/enum.rb
film42: I'll make an issue so we can get this fixed up.

film42:
This looks :+1: to me. @abrandoned / @mmmries ?

mmmries:
:+1: for me. I don't see any negative impacts

PR #266, opened 2015-09-25
resolve rubocop issues

mmmries:
https://twitter.com/madrubocop has been raining on the parade over in #264 

This PR is just to resolve all the rubocop errors so we don't have to disable the checks. Some of these changes I like and some I don't. I'll comment below.


lib/protobuf/decoder.rb
```diff
@@ -15,21 +15,21 @@ def self.decode_each_field(stream, &block)
...
-              when ::Protobuf::WireType::FIXED32 then
-                read_fixed32(stream)
-              when ::Protobuf::WireType::START_GROUP then
-                fail NotImplementedError, 'Group is deprecated.'
-              when ::Protobuf::WireType::END_GROUP then
-                fail NotImplementedError, 'Group is deprecated.'
-              else
-                fail InvalidWireType, wire_type
-              end
+      when ::Protobuf::WireType::VARINT then
```
mmmries: I preferred the way we were aligning these case statements previously. I think we might want to disable this cop rather than change the way we are indenting these.

ggilder:
It's probably a good idea to pin the version of Rubocop in the gemspec to avoid random build failures introduced by new checks...

mmmries:
@ggilder :+1: I went ahead and pegged it for now to avoid random failures in the future. I also change the Lint/EndAlignment to use the `keyword` mode since that seems to be the preferred style of the project.

ggilder:
:+1:

Issue #268, opened 2015-09-25
Use varint gem to speed up decoding?


zanker:
The protobuf gem https://github.com/codekitchen/ruby-protocol-buffers has a C extension called [varint](https://rubygems.org/gems/varint). If you take a look at a trace after the enum fix is applied

```
Measure Mode: wall_time
Thread ID: 70271059223040
Fiber ID: 70271070182440
Total: 3.308983
Sort by: self_time

 %self      total      self      wait     child     calls  name
 10.16      0.736     0.336     0.000     0.400   140000   <Class::Protobuf::Decoder>#read_varint
  4.92      0.182     0.163     0.000     0.019    94000   Kernel#respond_to?
  4.62      1.148     0.153     0.000     0.995    72000   <Class::Protobuf::Decoder>#read_field
  3.33      0.501     0.110     0.000     0.391    72000   <Class::Protobuf::Decoder>#read_key
  3.25      0.308     0.108     0.000     0.201    72000   Protobuf::Message::Fields::ClassMethods#get_field
  2.70      0.156     0.089     0.000     0.066   166500   IO::generic_readable#readbyte
  2.37      0.078     0.078     0.000     0.000   410500   Fixnum#&
  2.34      0.119     0.077     0.000     0.042   171500   Numeric#nonzero?
  2.29      0.076     0.076     0.000     0.000   144000   Protobuf::Field::BaseField#repeated?
  2.12      0.070     0.070     0.000     0.000    69000   Protobuf::Field::BaseField#setter
  2.05      0.068     0.068     0.000     0.000     1500   Kernel#caller
  2.00      0.066     0.066     0.000     0.000   166500   StringIO#getbyte
  1.69      0.056     0.056     0.000     0.000    72000   Protobuf::Message::Fields::ClassMethods#field_store
  1.53      0.093     0.050     0.000     0.042    72000   Protobuf::Field::BaseField#packed?
  1.41      0.102     0.047     0.000     0.055    51000   Kernel#dup
  1.37      0.045     0.045     0.000     0.000    47000   StringIO#read
  1.34      0.138     0.044     0.000     0.093    36003   Class#new
  1.34      0.285     0.044     0.000     0.241    43000   <Class::Protobuf::Decoder>#read_length_delimited
  1.34      0.165     0.044     0.000     0.121    30000   <Class::Protobuf::Enum>#fetch
  1.26      0.042     0.042     0.000     0.000   172000   Fixnum#zero?
[...]

PR #269, opened 2015-09-28
Varint optimizations

zanker:
Per #268, here's the results of benchmarking C vs pure Ruby. For the sake of being thorough, I also tested the variant used in the other protobuf gem (which was related to what @jjowdy) suggested, and it performed worst.

```
       user     system      total        real
pure_protobuf  1.080000   0.010000   1.090000 (  1.087074)
pure_alt  1.260000   0.010000   1.270000 (  1.275954)
c  0.700000   0.000000   0.700000 (  0.708581)
```

I also tested the C encoder in the varint gem, and it appears to be slower. This might be because with how protobuf is implemented, we need to create a `StringIO` class, rewind it, and then call `read` in order to get. Rewriting things might speed it up, but that's outside of what I'm looking to do in this PR.

Thanks @hollow and @codekitchen for the extension!

### DIFF
--- a/spec/lib/protobuf/varint_spec.rb
+++ b/spec/lib/protobuf/varint_spec.rb
@@ -1,0 +1,29 @@
+require 'base64'
+require 'spec_helper'
+
+RSpec.describe Protobuf::Varint do
+  VALUES = {
+    0 => "AA==",
+    5 => "BQ==",
+    51 => "Mw==",
+    9_192 => "6Ec=",
+    80_389 => "hfQE",
+    913_389 => "7d83",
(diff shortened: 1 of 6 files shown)


spec/lib/protobuf/varint_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'base64'
+require 'spec_helper'
+
+RSpec.describe Protobuf::Varint do
+  VALUES = {
```
zanker: I picked pretty arbitrary values to test that these are working. If someone has another suggestion, happy to include something else.

PR #271, opened 2015-09-28
Ensure ActiveSupport::Deprecation doesn't look for a caller when we h…

film42:
Big shout out to @brianstien for finding this problem.

This PR ensures that `ActiveSupport::Deprecation` doesn't look for a caller on deprecated fields when we have `PB_IGNORE_DEPRECATIONS` set to true. This turns out to be very expensive.

Here's profiling before this change:
```
 %self      total      self      wait     child     calls  name
 64.06      0.032     0.032     0.000     0.000     1000   Kernel#caller
  4.62      0.003     0.002     0.000     0.001     1000   Protobuf::FieldDeprecation#deprecated_method_warning
...
```

...and after:
```
 %self      total      self      wait     child     calls  name
 18.94      0.009     0.002     0.000     0.007     1000   Some::Protobuf::Model#some_deprecated_field
 18.10      0.002     0.002     0.000     0.001     1000   Protobuf::FieldDeprecation#deprecated_method_warning
  8.72      0.004     0.001     0.000     0.003     1000   ActiveSupport::Deprecation::Reporting#deprecation_warning
  8.05      0.005     0.001     0.000     0.004     1000   Protobuf::DeprecationBase#deprecation_warning
  7.42      0.001     0.001     0.000     0.000     1000   Kernel#tap
...
```

This turns out to be an order of magnitude faster for both MRI and Jruby:

...before:
```
irb(main):061:0> Benchmark.measure { 1000.times { t.some_deprecated_field } }.real.to_s
=> "0.3470001220703125"
```

...after:
```
irb(main):004:0> Benchmark.measure { 1000.times { t.some_deprecated_field  }}.real.to_s
=> "0.03399991989135742"
```

cc @abrandoned @brianstien @liveh2o @mmmries @localshred  @zanker


zanker:
Aha! I was curious where the `caller` ones were coming from. Nice find.

brianstien:
@zanker If you are using deprecated fields make sure you update both your clients and your servers, on messages with several deprecated fields it can be a pretty significant speedup.

PR #273, opened 2015-10-28
Remove GC start/stop around requests

film42:
This deprecates the `gc_pause_request` option from the CLI, and removes the `gc_pause` blocks around the process request methods. This may have been a speedup on MRI, but in a real threaded env, this functionality doesn't have the same guarantees. I vote we :fire: it.

cc @abrandoned @localshred @liveh2o @mmmries


localshred:
I think the deprecation notice is right, but seems odd to remove the behavior alongside the warning. Should we release a deprecation notice and when releasing 4.0 remove the feature?

film42:
@localshred That's probably a better solution.

film42:
@localshred done.

mmmries:
:+1: sounds like a good plan to me I vote we :shipit:

localshred:
WOO!! 

image

film42:
I can merge, but won't be able to cut a new version. Little help?

PR #275, opened 2015-11-16
forcing import requires to work

mooreniemi:
Using `protoc -I ./definitions --ruby_out ./lib/picatrix/generated definitions/*.proto` to generate my Ruby files, where on `.proto` file is importing another. Unless I make these changes, the generated code will have `require 'mason.pb'` which won't work. Should I be arranging my proto files differently to avoid this, or should some additional code be added to allow relative requiring?

### DIFF
--- a/lib/protobuf/generators/file_generator.rb
+++ b/lib/protobuf/generators/file_generator.rb
@@ -107,7 +107,8 @@ module Protobuf
         header "Imports"
 
         descriptor.dependency.each do |dependency|
-          print_require(convert_filename(dependency))
+          #print_require(convert_filename(dependency))
+          puts "'require_relative #{convert_filename(dependency)}'"
         end
 
         puts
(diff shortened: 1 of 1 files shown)


mmmries:
@mooreniemi at wrap this protobuf compilation in a gem so all of our `*.pb.rb` files are in the load path for our application.

So in this case if you add `./lib/picatrix/generated` to your ruby load path then it should be able to require `mason.pb` or `mason/jar.pb` without doing a relative require (note: I made up mason/jar.pb because I don't know anything about your message schema).

mooreniemi:
Ahhh thank you @mmmries that makes sense. Sorry for spamming you guys with questions this week!

mmmries:
:heart: :tada: :heart: :tada: :heart: :tada: I'm just happy that you are finding the library useful and it is helpful for me to hear about the parts that aren't obvious up-front.

Keep the questions coming

PR #277, opened 2016-01-18
Ensure the ping port is timed out after a reasonable amount of time

film42:
Should a server's ping port get jammed, overwhelmed, or access a port that is not open, then the ping port check will block until the TCPSocket default timeout rescues the blocked request. Instead, we can wrap Socket to add a reasonable timeout to ping port checks. By default, the timeout is set to 5 seconds, but you can customize this by setting the `PB_RPC_PING_PORT_TIMEOUT` environment variable.

There is probably some more work that can be done here, but I wanted this PR to be about solving this ping port problem before I make a few more changes, like using `PB_RPC_PING_PORT` inside of the new `Ping` class.

cc @abrandoned @quixoten @mmmries @liveh2o @brianstien @localshred

### DIFF
--- a/lib/protobuf/rpc/connectors/ping.rb
+++ b/lib/protobuf/rpc/connectors/ping.rb
@@ -1,0 +1,87 @@
+require "socket"
+
+module Protobuf
+  module Rpc
+    module Connectors
+      class Ping
+        attr_reader :host, :port
+
+        def initialize(host, port)
+          @host = host
+          @port = port
+        end
+
+        def online?
+          socket = tcp_socket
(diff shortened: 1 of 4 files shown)


mmmries:
Build failing looks like a bundler issue :dealwithbundler:

lib/protobuf/rpc/connectors/ping.rb
```diff
@@ -0,0 +1,87 @@
...
+        ensure
+          begin
+            socket && socket.close
+          rescue IOError
+            nil
+          end
+        end
+
+        def timeout
+          @timeout ||= begin
```
abrandoned: wouldn't we want the timeout to be in millis? (seconds seems like a long time)

film42: That's a good idea.

PR #278, opened 2016-01-20
Allow Message#each_field to return an enumerator

RKushnir:


mmmries:
This change looks good, but it looks like we are having bundler issues when building in travis. If you can get travis happy I'll gladly add my :+1: to this

RKushnir:
@mmmries Looks like someone was able to solve it by updating rubygems and bundler. 
https://github.com/pdfkit/pdfkit/pull/347
But I just googled it, cannot help more than that.

RKushnir:
I tried updating the bundler, but it still fails for jruby https://travis-ci.org/RKushnir/protobuf/builds/103723823

film42:
@RKushnir I just opened https://github.com/ruby-protobuf/protobuf/pull/279 which [makes travis pass](https://travis-ci.org/ruby-protobuf/protobuf/builds/103726222).

PR #279, opened 2016-01-20
Ensure bundler is at the latest version to make travis pass

film42:
Other github projects have seen the failing bundler issue and this is how they fixed it:

pdfkit/pdfkit#347
increments/qiita-markdown#24

Thanks @RKushnir for figuring it out.

cc @mmmries @brianstien @abrandoned @localshred

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -y libzmq3-dev
   - sudo -E ./install-protobuf.sh
+  - gem update bundler
 language: ruby
 rvm:
   - 1.9.3
--- a/spec/lib/protobuf_spec.rb
+++ b/spec/lib/protobuf_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe ::Protobuf do
     subject { ::Protobuf.client_host }
 
     context 'when client_host is not pre-configured' do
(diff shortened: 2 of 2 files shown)


mmmries:
@film42 looks like the travis logs are now available the only failure I am seeing is that our expectation for how `jruby` handles the `hostname` stuff is a bit off: https://travis-ci.org/ruby-protobuf/protobuf/jobs/103729876#L1024-L1030

mmmries:
@film42 I had to fix this in rspec_rabl with `gem install bundle -v 1.10.6` https://github.com/mmmries/rspec_rabl/blob/79c6c60d713e06dbba40a8d3dde5a8fd61bb34d7/.travis.yml#L11

The latest bundler just doesn't install the local gem correctly. Not sure why, but I downgraded to fix for now

film42:
@mmmries I just changed the spec to use ruby's `::Socket.gethostname` instead of the OS `hostname`, which was what Protobuf is doing. Comparing ruby vs the OS is a redundant test.

mmmries:
:+1: for not using the system `hostname` for our tests. That can obviously change based on which host the tests run on. I'l slightly concerned that we may have been expecting `Socket.hostname` to return only the last part of the fully-qualified hostname.

@localshred do you know if using the fully-qualified hostname from `Socket.hostname` will be a problem for us?

PR #280, opened 2016-01-25
WIP for discussion: Custom descriptor options

nerdrew:
This is a large PR with a ton of stuff jumbled together, but I have questions about how to slice it.

1) Objections to moving all the test protos to `spec/support/protos`? This would include `*.proto`, `*.pb.rb`, and `*.bin` files (proto file, generated code, and raw proto bytes).

2) 1 PR with a bunch of unrelated commits (e.g. 1 commit to cleanup `raise_error` warnings, 1 commit with an option to upcase enum names, 1 commit to move spec files, etc) vs a PR / commit?

3) Thoughts on adding a flag for generating gRPC specific service stubs?

Note: one reason this PR appears so large is that I refreshed some of the google unittest.proto files from upstream (will break that into either its own commit or its own PR pending question 2 above).

cc @lukaszx0


mmmries:
I think the best path forward is the one that keeps the potentially dangerous changes together so that the maintainers can focus on the PRs that matter most.

I suggest:
* 1 PR that just updates the upstream unit testing definitions and moves things into `spec/support`
  * it should be easy to see that this changes no behavior and only affects tests so we can verify it and get it merged right away
* 1 PR for each small behavior change
  * this will be a bit more work for you, but makes it a lot more likely that your PRs will get looked at and merged quickly.
  * A PR that just cleans up `raise_error` warnings is easy to reason about and ship

nerdrew:
Sounds Good. I'm going to trickle the PRs in so I don't have to rebase all over the place: https://github.com/ruby-protobuf/protobuf/pull/281 (fix rspec warnings)

nerdrew:
Fix for `PB_NO_TAG_WARNINGS` pollution: https://github.com/ruby-protobuf/protobuf/pull/282

nerdrew:
Consolidate, cleanup, and refresh protos from upstream: https://github.com/ruby-protobuf/protobuf/pull/283

nerdrew:
Add missing deprecation requires: https://github.com/ruby-protobuf/protobuf/pull/284

nerdrew:
First feature PR, add `PB_ENUM_UPCASE` option: https://github.com/ruby-protobuf/protobuf/pull/285

nerdrew:
Closing this PR for now. Will open a new PR for the custom options / extensions after the others are merged.

PR #281, opened 2016-01-27
fix raise_error rspec warnings

nerdrew:


zachmargolis:
LGTM!

mmmries:
Any insight as to why the travis build is failing? The latest successful master build was a while ago, but they both seem to be using bundler `1.7.6`.

Maybe you need to add a `before_install` step to update the bundler version?

PR #283, opened 2016-01-27
consolidate test protos to spec/support/protos

nerdrew:
- add SUPPORT_PATH, PROTOS_PATH helpers in specs
- refresh unittest protos from upstream


spec/support/protos/google_unittest.proto
```diff
@@ -663,26 +829,24 @@ message TestParsingMerge {
...
+    repeated group Group2 = 20 {
+      optional TestAllTypes field1 = 21;
+    }
     repeated TestAllTypes ext1 = 1000;
     repeated TestAllTypes ext2 = 1001;
   }
   required TestAllTypes required_all_types = 1;
   optional TestAllTypes optional_all_types = 2;
   repeated TestAllTypes repeated_all_types = 3;
-  // TODO: Support groups
```
zachmargolis: are groups supported? or should this remain commented out?

nerdrew: Good catch

nerdrew: Hmmm. Strange. It is able to read the proto and the generated code looks right. Lemme dig up the protobuf docs on group.

Leaving it as is. It seems to work fine.

zachmargolis: Cool :sunglasses:

PR #285, opened 2016-01-27
[NEW FEATURE] add PB_ENUM_UPCASE option

nerdrew:
enum names are not required to be Capitalized, but ruby constants are.
Add option to UPCASE all enum names

This proto definition:

```
enum SillyEnum {
  foo = 1;
  BAR = 2;
  BAZ = 3;
}
```

Generates this code:

```
class SillyEnum < ::Protobuf::Enum
  define :foo, 1
  define :BAR, 2
  define :BAZ, 3
end
```

I get this error loading the generated code:

```
% pry -I. -r protobuf/deprecation -r silly_enum.pb
.../lib/protobuf/enum.rb:58:in `const_set': wrong constant name foo (NameError)
	from .../lib/protobuf/enum.rb:58:in `define'
```

Enabling `PB_ENUM_UPCASE` changes the generated code to:

```
class SillyEnum < ::Protobuf::Enum
  define :FOO, 1
  define :BAR, 2
  define :BAZ, 3
end
```

### DIFF
--- a/lib/protobuf/generators/enum_generator.rb
+++ b/lib/protobuf/generators/enum_generator.rb
@@ -32,6 +32,7 @@ module Protobuf
 
       def build_value(enum_value_descriptor)
         name = enum_value_descriptor.name
+        name.upcase! if ENV.key?('PB_UPCASE_ENUMS')
         number = enum_value_descriptor.number
         "define :#{name}, #{number}"
       end
--- a/spec/lib/protobuf/generators/enum_generator_spec.rb
+++ b/spec/lib/protobuf/generators/enum_generator_spec.rb
@@ -68,6 +68,15 @@ end
     it 'returns a string identifying the given enum value' do
       expect(subject.build_value(enum.value.first)).to eq("define :FOO, 1")
     end
+
+    context 'with PB_UPCASE_ENUMS set' do
+      before { allow(ENV).to receive(:key?).with('PB_UPCASE_ENUMS').and_return(true) }
+      let(:values) { [{ :name => 'boom', :number => 1 }] }
+
+      it 'returns a string with the given enum name in ALL CAPS' do
+        expect(subject.build_value(enum.value.first)).to eq("define :BOOM, 1")
+      end
+    end
   end
 
 end


lib/protobuf/generators/enum_generator.rb
```diff
@@ -32,6 +32,7 @@ def compile
 
       def build_value(enum_value_descriptor)
         name = enum_value_descriptor.name
+        name.upcase! if ENV.fetch('PB_ENUM_UPCASE', '') != ''
```
zachmargolis: what about `ENV['PB_ENUM_UPCASE'].present?` (activesupport is already a dependency of this gem, so we could just add `require 'activesupport/core_ext/object/blank'`)

zachmargolis:
Big fan of this. The only problem is that the string names would be different (ex if transporting these values over JSON), but this is the simplest way to get valid Ruby code

lib/protobuf/generators/enum_generator.rb
nerdrew: I'm going to switch it to `key?` to match the other usages

SalvatoreT:
:+1:

nerdrew:
Any comments on this?

nerdrew:
@embark @film42 How does this look?

embark:
:+1: I think I'm okay with the string/json problem -- 
1) this is an opt-in feature and 
2) if anyone using this gem relied on non-uppercase enums AND for some reason were passing protos around as json, the gem wouldn't be working for them right now anyway.

liveh2o:
We can certainly handle this as implemented here, but I'm wondering if this is really a problem that we should be handling by upcasing the values in the protobuf definition. I can't see anything in the docs that requires enum names to be uppercase, but every example I can find (even in the [generated code guide](https://developers.google.com/protocol-buffers/docs/reference/overview)) has enum names defined in uppercase.

What about adding an option that will bail during compilation if enums are not uppercased? That seems to fit with the current behavior where enums with overlapping names in the same namespace (i.e.

```
enum SillyEnum {
  FOO = 1;
  BAR = 2;
  BAZ = 3;
}

enum SillierEnum {
  FOO = 1;
  BAR = 2;
  BAZ = 3;
}
```

) will fail because enumerations within the same namespace must be uniq in C++. I'm thinking it would be something like `PB_REQUIRE_UPPERCASED_ENUMS=1`. Compiling with that option would require protobuf definitions to use uppercased enum names.

nerdrew:
Ya, agreed on enums *should* always be upcased, but we have tons of enums that are lowercase or CamelCase. Not sure why, but we can't change it now. We currently upcase all enum names in our fork of them gem.

lib/protobuf/generators/enum_generator.rb
```diff
@@ -32,6 +32,7 @@ def compile
 
       def build_value(enum_value_descriptor)
         name = enum_value_descriptor.name
+        name.upcase! if ENV.key?('PB_ENUM_UPCASE')
```
film42: Thank you! One change recommendation and I'll merge this. Can we change this from `PB_ENUM_UPCASE ` to `PB_UPCASE_ENUMS`? That way it fits with the language of `PB_IGNORE_DEPRECATIONS` and other env variables.

nerdrew: Renamed.

film42:
:shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit:

film42:
I'll work on cutting a pre-release soon.

PR #286, opened 2016-02-01
only fetch value once on serialization and cache the varint encoding …

abrandoned:
…for most common cases (< 1024)

adding a cache of varint fields for any varint < 1024 for now (may make configurable in the future), runs it on load and uses the cached values instead of calculating

https://gist.github.com/abrandoned/044cf200037ff7885572

about 100% increase in encoding time of varints under jruby when varint is > 128; < 128 10%

### DIFF
--- a/lib/protobuf/field/varint_field.rb
+++ b/lib/protobuf/field/varint_field.rb
@@ -8,6 +8,7 @@ module Protobuf
       # Constants
       #
 
+      CACHE_LIMIT = 1024
       INT32_MAX  =  2**31 - 1
       INT32_MIN  = -2**31
       INT64_MAX  =  2**63 - 1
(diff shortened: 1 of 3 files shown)


lib/protobuf/field/varint_field.rb
```diff
@@ -32,6 +42,11 @@ def self.encode(value)
         (bytes << value).pack('C*')
       end
 
+      # Load the cache of VarInts on load of file
+      (0..CACHE_LIMIT).to_a.each do |cached_value|
```
zachmargolis: maybe a little late, but ranges are iterable without needing the `.to_a`, the `.to_a` ends up creating a medium-sized array that's not really used

abrandoned: @zachmargolis we can include the update in the next release (I will be reviewing the PR's submitted today, tonight)

it only runs during loading of the file, so shouldn't have a running impact

PR #292, opened 2016-02-03
Abrandoned/serialization encoding optimizations

abrandoned:
"larger" change with removing `new().encode` from encoder, creates lot of new objects and garbage during serialization without any noticeable benefit (removing took 1 second off of 14 second 100_000 record serialization)

ordering of `nil?` `is_a?` checks in string instances favored `nil?` case, but `String` case is "more common" with deserialization (as a string field would not be deserialized unless it was present during serialization) <= reordered acceptability check

added caching of varint => 2048 by default because of how wire types serialize; the shift makes 1049 a common "tag" (if > 128 fields)

removed dual varint encoders as class level will suffice (and instance level weren't cached)

removed multiple field stores in "fields" as the symbol will not negatively interact with the string and the tag; having Ruby only calculate the hash key once reduces lookup times for fields during serialization

More work to be done, but will put other changes into issues that are reported (around inconsistent serialization and deserialization concerns); looking to keep this one focused on maintaining all current behavior

@film42 @zachmargolis @mmmries

### DIFF
--- a/lib/protobuf/field/bool_field.rb
+++ b/lib/protobuf/field/bool_field.rb
@@ -3,6 +3,8 @@ require 'protobuf/field/varint_field'
 module Protobuf
   module Field
     class BoolField < VarintField
+      FALSE_STRING = "false".freeze
+      TRUE_STRING = "true".freeze
 
       ##
       # Class Methods
@@ -23,9 +25,9 @@ module Protobuf
       def coerce!(val)
         case val
         when String then
-          if val == 'true'
+          if val == TRUE_STRING
             true
-          elsif val == 'false'
+          else # acceptable? is called before coerce, so we don't need to check for now
             false
           end
         else
--- a/lib/protobuf/field/bool_field.rb
+++ b/lib/protobuf/field/bool_field.rb
@@ -21,10 +21,13 @@ module Protobuf
       end
 
       def coerce!(val)
-        if val == 'true'
-          true
-        elsif val == 'false'
-          false
+        case val
+        when String then
+          if val == 'true'
+            true
+          elsif val == 'false'
+            false
+          end
         else
           val
         end
--- a/lib/protobuf/field/bool_field.rb
+++ b/lib/protobuf/field/bool_field.rb
@@ -24,12 +24,8 @@ module Protobuf
 
       def coerce!(val)
         case val
-        when String then
-          if val == TRUE_STRING
-            true
-          else # acceptable? is called before coerce, so we don't need to check for now
-            false
-          end
+        when String
+          val == TRUE_STRING
         else
           val
         end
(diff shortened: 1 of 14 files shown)


abrandoned:
@brianstien

lib/protobuf/field/bool_field.rb
```diff
@@ -21,10 +21,13 @@ def acceptable?(val)
...
       def coerce!(val)
-        if val == 'true'
-          true
-        elsif val == 'false'
-          false
+        case val
+        when String then
+          if val == 'true'
+            true
+          elsif val == 'false'
```
liveh2o: Perhaps we can move these to constants so we aren't creating new string objects purely for checking for true/false here.

lib/protobuf/field/message_field.rb
```diff
@@ -9,11 +9,7 @@ class MessageField < BaseField
       #
 
       def acceptable?(val)
-        unless val.is_a?(type_class) || val.respond_to?(:to_hash)
-          fail TypeError, "Expected value of type '#{type_class}' for field #{name}, but got '#{val.class}'"
```
liveh2o: I think this is a good change, but wanted to call this out as a behavior change since you mentioned you were trying to maintain existing behavior.

liveh2o:
Overall, a good set of changes. Glad to see improved serialization here.

lib/protobuf/field/message_field.rb
abrandoned: it's not a behavior change; we do the same thing in the serialization layer when calling `acceptable?(val)`; if it isn't acceptable we fail with the message there (duplication of behavior)

lib/protobuf/field/bytes_field.rb
```diff
@@ -58,11 +58,12 @@ def define_setter
         message_class.class_eval do
           define_method(method_name) do |val|
             begin
-              val = "#{val}" if val.is_a?(Symbol)
-
-              if val.nil?
+              case val
+              when String, Symbol then
```
zachmargolis: style q: why use `then` for a multiline case? I'm only used to using it for single-line

lib/protobuf/field/varint_field.rb
```diff
@@ -43,7 +43,7 @@ def self.encode(value, use_cache = true)
       end
 
       # Load the cache of VarInts on load of file
-      (0..CACHE_LIMIT).to_a.each do |cached_value|
+      (0..CACHE_LIMIT).each do |cached_value|
```
zachmargolis: :heart_eyes: thx for this small fix

lib/protobuf/field/bytes_field.rb
abrandoned: will update that, habit

film42:
:shipit:

PR #293, opened 2016-02-05
Speed Up Creates & Updates By Memoizing on encode

vintagepenguin:
Serialization is one of the slowest things that you do in Protobuf. If we already have an encoded message, then we don't need to encode it again so why not memoize it. 

@abrandoned @liveh2o @mmmries @brianstien

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -194,7 +194,7 @@ module Protobuf
 
         message_class.class_eval do
           define_method(method_name) do |val|
-            @memoized_encoded = nil
+            @encode = nil
             if val.is_a?(Array)
               val = val.dup
               val.compact!
@@ -236,7 +236,7 @@ module Protobuf
 
         message_class.class_eval do
           define_method(method_name) do |val|
-            @memoized_encoded = nil
+            @encode = nil
             if val.nil? || (val.respond_to?(:empty?) && val.empty?)
               @values.delete(field.name)
             elsif field.acceptable?(val)
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -21,7 +21,6 @@ module Protobuf
       ##
       # Attributes
       #
-
       attr_reader :message_class, :name, :options, :rule, :tag, :type_class
 
       ##
@@ -195,6 +194,7 @@ module Protobuf
 
         message_class.class_eval do
           define_method(method_name) do |val|
+            @memoized_encoded = nil
             if val.is_a?(Array)
               val = val.dup
               val.compact!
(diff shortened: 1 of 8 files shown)


lib/protobuf/message.rb
```diff
@@ -27,6 +27,11 @@ class Message
     include ::Protobuf::Message::Serialization
 
     ##
+    # Attributes
+    #
+    attr_accessor :memoized_encoded
```
abrandoned: why would we need the attr_accessor? (either of them)

lib/protobuf/message/serialization.rb
```diff
@@ -48,10 +48,12 @@ def decode_from(stream)
       # Encode this message
       #
       def encode
-        stream = ::StringIO.new
-        stream.set_encoding(::Protobuf::Field::BytesField::BYTES_ENCODING)
-        encode_to(stream)
-        stream.string
+        @memoized_encoded ||= begin
```
abrandoned: if the value is going to be memoized then we should call the instance var the same as the method

lib/protobuf/field/base_field.rb
```diff
@@ -21,7 +21,7 @@ class BaseField
       ##
       # Attributes
       #
-
+      attr_accessor :memoized_encoded
```
abrandoned: again, not sure why this is necessary

lib/protobuf/message/serialization.rb
film42: I think it would be nice to read...

```
def encode
  @encode = nil if @has_changed
  @encode ||= begin
    ...
  end
end
```

lib/protobuf/field/base_field.rb
```diff
@@ -195,6 +195,7 @@ def define_array_setter
 
         message_class.class_eval do
           define_method(method_name) do |val|
+            @memoized_encoded = nil
```
film42: I think we might want to say `@has_change = true`. It becomes more clear to me what we're trying to keep track of. That's just my opinion though.

vintagepenguin:
@film42 commenting here since the diff you commented on got deleted. 

But yeah, I think `@encode = nil` it is confusing if you're looking at this for sure. 

But the goal here it to clear out that instance variable so that when we hit the memoization in the `encode` method we know that we need to `encode` the message again because there has been a change and we know that because the `setter` was called.

film42:
@vintagepenguin It would be more clear set a flag, or call a `bust_encode_cache!` method or something. But this seems :+1: . 

Looks like you have a :cop: to fix, but then :shipit: .

goddardc:
It seems that with this change, performing operations on repeated fields is no longer handled properly after the memoized encoding has been computed.

```ruby
a.encode
a.some_repeated_field << "new-value"
a.encode # doesn't contain "new-value"
```

liveh2o:
That's definitely a problem, @goddardc. Mind writing up an issue?

goddardc:
@liveh2o Sure thing!

PR #295, opened 2016-02-06
already loading field on define and should not need to calcualte the …

abrandoned:
…tag encoding at runtime

@film42 per our discussion on knowing the tag Varint at load time; we can store the encoded tag value on the field object in the definition and read it directly during the serialization phase instead of calculating the varint again

removes ~1.5 seconds on 15 second 100_000 object serialize

@film42

### DIFF
--- a/lib/protobuf/field/bytes_field.rb
+++ b/lib/protobuf/field/bytes_field.rb
@@ -58,21 +58,15 @@ module Protobuf
         message_class.class_eval do
           define_method(method_name) do |val|
             @encode = nil
-            begin
-              case val
-              when String, Symbol
-                @values[field.name] = "#{val}"
-              when NilClass
-                @values.delete(field.name)
-              when ::Protobuf::Message
-                @values[field.name] = val.dup
-              else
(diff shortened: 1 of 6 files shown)


lib/protobuf/field/bytes_field.rb
```diff
@@ -58,21 +58,15 @@ def define_setter
...
-              when String, Symbol
-                @values[field.name] = "#{val}"
-              when NilClass
-                @values.delete(field.name)
-              when ::Protobuf::Message
-                @values[field.name] = val.dup
-              else
-                fail TypeError, "Unacceptable value #{val} for field #{field.name} of type #{field.type_class}"
-              end
-            rescue NoMethodError => ex
```
mmmries: Interesting that we had no test coverage for this behavior. In either case we are raising an exception and we can let the caller decider if they want to log and convert the error type themselves :+1:

mmmries:
If all the travis checks pass I say :shipit:

film42:
:+1: :+1:

PR #296, opened 2016-02-07
compress decoder to single method; only check nil? once on serializat…

abrandoned:
…ion during required? check; add decoding optimization for enum values in "fetch"

small optimizations towards a faster decoder

@film42


film42:
:+1:

PR #297, opened 2016-02-08
Abrandoned/decoding setters without copy

abrandoned:
start moving towards per field decoding setters instead of using the standard setter during the decode routine; the standard setter does copies to keep from mutating an incoming value when a value is set, during decode we also make a copy as a result of pulling it off the stream, largely reduces a single copy and in cases where we can it prevents the acceptable? checks as a value that has been serialized and is tagged should be acceptable

also started added setters directly on the message object by tag instead of doing a field lookup and then running through the name of the field; reduces the amount of searching for field definitions and will eventually lead to us having a deserialization routine that only utilizes tags (instead of field lookup)

on MRI and JRuby these changes accounted for ~20% decrease in time for serialization/deserialization of the command
`bx rake benchmark:profile_protobuf_serialize[100000,prof_out.txt]`

@film42

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -176,6 +176,7 @@ module Protobuf
         else
           define_getter
           define_setter
+          define_decode_setter
         end
       end
 
@@ -232,6 +233,10 @@ module Protobuf
         end
 
         ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
+      end
+
+      def define_decode_setter
+        # empty for now
       end
 
       def define_setter
--- a/lib/protobuf/field/bool_field.rb
+++ b/lib/protobuf/field/bool_field.rb
@@ -35,6 +35,20 @@ module Protobuf
         value == 1
       end
 
+      def define_decode_setter
+        field = self
+        name_method_name = "_protobuf_decode_setter_#{field.name}"
+        tag_method_name = "_protobuf_decode_setter_#{field.tag}"
+
+        message_class.class_eval do
+          define_method(name_method_name) do |val|
+            @values[field.name] = field.decode(val)
+          end
+
+          alias_method tag_method_name, name_method_name
+        end
+      end
+
       def encode(value)
         [value ? 1 : 0].pack('C')
       end
--- a/lib/protobuf/field/varint_field.rb
+++ b/lib/protobuf/field/varint_field.rb
@@ -63,6 +63,20 @@ module Protobuf
         Integer(val, 10)
       end
 
+      def define_decode_setter
+        field = self
+        name_method_name = "_protobuf_decode_setter_#{field.name}"
+        tag_method_name = "_protobuf_decode_setter_#{field.tag}"
+
+        message_class.class_eval do
+          define_method(name_method_name) do |val|
+            @values[field.name] = val
+          end
+
+          alias_method tag_method_name, name_method_name
+        end
+      end
+
       def decode(value)
         value
       end
--- a/lib/protobuf/field/varint_field.rb
+++ b/lib/protobuf/field/varint_field.rb
@@ -52,6 +52,7 @@ module Protobuf
       #
 
       def acceptable?(val)
+        return true if val.is_a?(Integer) && val >= 0 && val < INT32_MAX
         int_val = coerce!(val)
         int_val >= self.class.min && int_val <= self.class.max
       rescue
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -77,8 +77,7 @@ module Protobuf
       return to_enum(:each_field) unless block_given?
 
       self.class.all_fields.each do |field|
-        value = __send__(field.getter)
-        yield(field, value)
+        yield(field, __send__(field.getter))
       end
     end
 
--- a/lib/protobuf/field/bytes_field.rb
+++ b/lib/protobuf/field/bytes_field.rb
@@ -56,6 +56,21 @@ module Protobuf
       # Private Instance Methods
       #
 
+      def define_decode_setter
+        field = self
+        name_method_name = "_protobuf_decode_setter_#{field.name}"
+        tag_method_name = "_protobuf_decode_setter_#{field.tag}"
+
+        message_class.class_eval do
+          define_method(name_method_name) do |val|
+            @encode = nil
+            @values[field.name] = field.decode(val)
+          end
+
+          alias_method tag_method_name, name_method_name
+        end
+      end
+
       def define_setter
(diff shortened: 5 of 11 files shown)


lib/protobuf/field/base_field.rb
```diff
@@ -234,6 +235,10 @@ def define_getter
         ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
       end
 
+      def define_decode_setter
+        # empty for now
```
film42: Only a few variants of this method. Let's bring the duplicated common implementation here.

abrandoned: probably best, I wrote this while I was testing the string copy only and now need to refactor since a common pattern has emerged

film42: :+1:

abrandoned:
@zachmargolis would love to get your thoughts on this path

lib/protobuf/field/varint_field.rb
```diff
@@ -52,6 +52,7 @@ def self.encode(value, use_cache = true)
       #
 
       def acceptable?(val)
+        return true if val.is_a?(Integer) && val >= 0 && val < INT32_MAX
```
zachmargolis: we still want to compare to `self.class.max` right?

I understand wanting to avoid `coerce!` for performance reasons but maybe there's a better way to preserve the existing logic?

Just an idea

```ruby
def acceptable?(val)
  int_val = val.is_a?(Integer)
    val
  else
    coerce!(val)
  end

  int_val >= self.class.min && int_val <= self.class.max
end

```

lib/protobuf/field/bool_field.rb
```diff
@@ -35,6 +35,21 @@ def decode(value)
...
+        field_name = field.name
+        name_method_name = "_protobuf_decode_setter_#{field.name}"
+        tag_method_name = "_protobuf_decode_setter_#{field.tag}"
+
+        message_class.class_eval do
+          define_method(name_method_name) do |val|
+            @values[field_name] = field.decode(val)
+          end
+
+          alias_method tag_method_name, name_method_name
```
zachmargolis: in `lib/protobuf/message/serialization.rb` looks like we only ever call the tag version of this method, why even define the field name versions at all?

lib/protobuf/field/varint_field.rb
abrandoned: we do want to compare the max (which happens after this if it is larger than the smallest max); the smallest max is the INT32_MAX, so if it passes it (and is an Integer; which is the most common case, especially in deserialization) then we just set it instead of taking the long route (which in our usage is very uncommon)

lib/protobuf/field/bool_field.rb
abrandoned: I think we can take out the name versions; I was originally just looking for a tag version so decoding doesn't need to lookup the field; will change this

lib/protobuf/field/varint_field.rb
zachmargolis: I guess as a reader of this patch, I'm having a hard following what it's supposed to do. I didn't realize it would fall through to the old case. Maybe we could do something like this (comments included) for posterity?

```ruby
def acceptable?(val)
  int_val = if val.is_a?(Integer)
    return true if val >= 0 && val < INT32_MAX # return quickly for smallest integer size, hot code path
    val
  else
    coerce!(val)
  end
  int_val >= self.class.min && int_val <= self.class.max
end
```

lib/protobuf/field/bool_field.rb
zachmargolis: Awesome, I think that this will help avoid some extension vs field issues later down the line as well, since tags are 100% unambiguous

lib/protobuf/field/base_field.rb
```diff
@@ -176,16 +176,18 @@ def define_accessor
...
 
       def define_array_getter
         field = self
+        field_name = field.name
         method_name = field.getter
 
         message_class.class_eval do
           define_method(method_name) do
-            @values[field.name] ||= ::Protobuf::Field::FieldArray.new(field)
+            @values[field_name] ||= ::Protobuf::Field::FieldArray.new(field)
```
embark: Heads up that a lot of this work will need to be moved to the `[]=` and `[]` methods after PR #302 is merged. In fact a lot of the setter/getter code in this PR will conflict, I comment on just a little of it below.

lib/protobuf/field/bool_field.rb
```diff
@@ -35,6 +35,19 @@ def decode(value)
...
       end
 
+      def define_decode_setter
+        field = self
+        field_name = field.name
+        tag_method_name = "_protobuf_decode_setter_#{field.tag}"
+
+        message_class.class_eval do
+          define_method(tag_method_name) do |val|
+            @encode = nil
```
embark: I think `@encode` has gone away, from what @film42 has commented.

lib/protobuf/message.rb
```diff
@@ -77,14 +77,13 @@ def each_field
       return to_enum(:each_field) unless block_given?
 
       self.class.all_fields.each do |field|
-        value = __send__(field.getter)
-        yield(field, value)
+        yield(field, __send__(field.getter))
```
embark: After #302 it would be the style to use `yield(field, self[field.name])`, you'll probably get a merge conflict anyway, and in `each_field_for_serialization` it would probably be `self[field.name]` too.

lib/protobuf/varint_pure.rb
```diff
@@ -1,12 +1,12 @@
 module Protobuf
   module VarintPure
     def decode(stream)
-      value = index = 0
+      value = shift = 0
       begin
         byte = stream.readbyte
-        value |= (byte & 0x7f) << (7 * index)
```
embark: Bikeshedding a little, but honestly I prefer the hex. Hex makes it easier to picture the individual bits in my head.

lib/protobuf/field/bool_field.rb
```diff
@@ -35,6 +35,19 @@ def decode(value)
         value == 1
       end
 
+      def define_decode_setter
+        field = self
+        field_name = field.name
+        tag_method_name = "_protobuf_decode_setter_#{field.tag}"
```
embark: Can we make this private, `Message.instance_eval { private tag_method_name.to_sym }` or some such?

embark:
:+1: great optimization!

How much benefit do we get on creating the dynamic per-tag methods vs. having 1 method in the message where that requires a field lookup first? 

Just curious if the optimization is worth the complexity tradeoff of a bunch of new dynamically generated methods

PR #299, opened 2016-02-08
Fix #define_getter method

akihiro17:
If a required bool field is set to `false`, `#define_getter` returns its default value(=nil) wrongly because of the following code.

https://github.com/ruby-protobuf/protobuf/blob/master/lib/protobuf/field/base_field.rb#L230
```ruby
def define_getter
---
  @values[field.name] || field.default_value
```

This behavor was introduced by [this commit](https://github.com/ruby-protobuf/protobuf/commit/05aa7756e489e792c7bb0785ef2979931a4879a6)

This commit fixes the above issue

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -227,7 +227,7 @@ module Protobuf
 
         message_class.class_eval do
           define_method(method_name) do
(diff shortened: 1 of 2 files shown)


film42:
:+1: Good catch. Thank you!

PR #302, opened 2016-02-12
Clean up dynamic code generation in prep for extension namespacing

embark:
CC @nerdrew

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -35,7 +35,7 @@ module Protobuf
       # Constructor
       #
 
-      def initialize(message_class, rule, type_class, name, tag, options)
+      def initialize(message_class, rule, type_class, name, tag, simple_name, options)
         @message_class = message_class
         @name          = name
         @rule          = rule
@@ -44,7 +44,7 @@ module Protobuf
         @options       = options
 
         validate_packed_field if packed?
-        define_accessor
+        define_accessor(simple_name, name) if simple_name
         tag_encoded
       end
 
@@ -69,10 +69,12 @@ module Protobuf
       end
 
       def default_value
-        @default_value ||= case
-                           when optional? then typed_default_value
-                           when repeated? then ::Protobuf::Field::FieldArray.new(self).freeze
-                           when required? then nil
+        @default_value ||= if optional? || required?
+                             typed_default_value
+                           elsif repeated?
+                             ::Protobuf::Field::FieldArray.new(self).freeze
+                           else
+                             fail "Unknown field label -- something went very wrong"
                            end
       end
 
@@ -122,10 +124,10 @@ module Protobuf
 
       # FIXME: need to cleanup (rename) this warthog of a method.
       def set(message_instance, bytes)
-        return message_instance.__send__(setter, decode(bytes)) unless repeated?
-        return message_instance.__send__(getter) << decode(bytes) unless packed?
+        return message_instance[name] = decode(bytes) unless repeated?
+        return message_instance[name] << decode(bytes) unless packed?
 
-        array = message_instance.__send__(getter)
+        array = message_instance[name]
         stream = StringIO.new(bytes)
 
         if wire_type == ::Protobuf::WireType::VARINT
@@ -169,87 +171,23 @@ module Protobuf
       # Private Instance Methods
       #
 
-      def define_accessor
-        if repeated?
-          define_array_getter
-          define_array_setter
-        else
-          define_getter
-          define_setter
-        end
-      end
-
-      def define_array_getter
-        field = self
-        method_name = field.getter
-
+      def define_accessor(simple_field_name, fully_qualified_field_name)
         message_class.class_eval do
-          define_method(method_name) do
-            @values[field.name] ||= ::Protobuf::Field::FieldArray.new(field)
+          define_method("#{simple_field_name}!") do
+            @values[fully_qualified_field_name]
           end
         end
 
-        ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
-      end
-
-      def define_array_setter
-        field = self
-        method_name = field.setter
-
         message_class.class_eval do
-          define_method(method_name) do |val|
-            if val.is_a?(Array)
-              val = val.dup
-              val.compact!
-            else
-              fail TypeError, <<-TYPE_ERROR
-                Expected repeated value of type '#{field.type_class}'
-                Got '#{val.class}' for repeated protobuf field #{field.name}
-              TYPE_ERROR
-            end
-
-            if val.nil? || (val.respond_to?(:empty?) && val.empty?)
-              @values.delete(field.name)
-            else
-              @values[field.name] ||= ::Protobuf::Field::FieldArray.new(field)
-              @values[field.name].replace(val)
-            end
-          end
+          define_method(simple_field_name) { self[fully_qualified_field_name] }
+          define_method("#{simple_field_name}=") { |v| self[fully_qualified_field_name] = v }
         end
 
-        ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
-      end
+        return unless deprecated?
 
-      def define_getter
-        field = self
-        method_name = field.getter
-
-        message_class.class_eval do
-          define_method(method_name) do
-            @values[field.name] || field.default_value
-          end
-        end
-
-        ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
-      end
-
-      def define_setter
-        field = self
-        method_name = field.setter
-
-        message_class.class_eval do
-          define_method(method_name) do |val|
-            if val.nil? || (val.respond_to?(:empty?) && val.empty?)
-              @values.delete(field.name)
-            elsif field.acceptable?(val)
-              @values[field.name] = field.coerce!(val)
-            else
-              fail TypeError, "Unacceptable value #{val} for field #{field.name} of type #{field.type_class}"
-            end
-          end
-        end
-
-        ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
+        ::Protobuf.field_deprecator.deprecate_method(message_class, simple_field_name)
+        ::Protobuf.field_deprecator.deprecate_method(message_class, "#{simple_field_name}!")
+        ::Protobuf.field_deprecator.deprecate_method(message_class, "#{simple_field_name}=")
       end
 
       def typed_default_value
--- a/lib/protobuf/field/bytes_field.rb
+++ b/lib/protobuf/field/bytes_field.rb
@@ -33,7 +33,6 @@ module Protobuf
       end
 
       def encode(value)
-        value_to_encode = ""
         if value.is_a?(::Protobuf::Message)
           value_to_encode = value.encode
         else
@@ -50,32 +49,17 @@ module Protobuf
         ::Protobuf::WireType::LENGTH_DELIMITED
       end
 
-      private
-
-      ##
-      # Private Instance Methods
-      #
-
-      def define_setter
-        field = self
-        method_name = field.setter
-
-        message_class.class_eval do
-          define_method(method_name) do |val|
-            case val
-            when String, Symbol
-              @values[field.name] = "#{val}"
-            when NilClass
-              @values.delete(field.name)
-            when ::Protobuf::Message
-              @values[field.name] = val.dup
-            else
-              fail TypeError, "Unacceptable value #{val} for field #{field.name} of type #{field.type_class}"
-            end
-          end
+      def coerce!(value)
+        case value
+        when String, Symbol
+          "#{value}"
+        when NilClass
+          nil
+        when ::Protobuf::Message
+          value.dup
+        else
+          fail TypeError, "Unacceptable value #{value} for field #{name} of type #{type_class}"
         end
-
-        ::Protobuf.field_deprecator.deprecate_method(message_class, method_name) if field.deprecated?
       end
     end
   end
--- a/lib/protobuf/field/enum_field.rb
+++ b/lib/protobuf/field/enum_field.rb
@@ -32,28 +32,17 @@ module Protobuf
         true
       end
 
+      def coerce!(value)
+        enum_value = type_class.fetch(value)
+        fail TypeError, "Invalid Enum value: #{value.inspect} for #{name}" unless enum_value
+        enum_value
+      end
+
       private
 
       ##
       # Private Instance Methods
       #
-
-      def define_setter
-        field = self
-        message_class.class_eval do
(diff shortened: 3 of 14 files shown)


lib/protobuf/message.rb
```diff
@@ -140,20 +140,54 @@ def to_proto
...
     end
 
     def []=(name, value)
       if (field = self.class.get_field(name, true))
-        __send__(field.setter, value) unless value.nil?
+        @encode = nil
+        if field.repeated?
+          if value.is_a?(Array)
+            value = value.dup
+            value.compact!
```
zachmargolis: I know this is preserving existing behavior, but just want to note that IMO this `compact` thing is super weird behavior, settings an array with nil values in there should def be an error

spec/lib/protobuf/field/bool_field_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.describe Protobuf::Field::Int32Field do
+RSpec.describe Protobuf::Field::BoolField do
```
zachmargolis: good catch :sunglasses:

lib/protobuf/field/base_field.rb
```diff
@@ -169,89 +171,23 @@ def wire_type
...
-        field = self
-        method_name = field.getter
-
-        message_class.class_eval do
-          define_method(method_name) do
-            @values[field.name] || field.default_value
-          end
+        if deprecated?
+          ::Protobuf.field_deprecator.deprecate_method(message_class, simple_field_name)
+          ::Protobuf.field_deprecator.deprecate_method(message_class, "#{simple_field_name}!")
```
zachmargolis: would it make sense to deprecate the `?` method as well too?

zachmargolis:
Should we add a test that verified that the fully qualified names are in `@values` to make sure they are set correctly?

lib/protobuf/field/base_field.rb
```diff
@@ -69,10 +69,12 @@ def default
       end
 
       def default_value
-        @default_value ||= case
-                           when optional? then typed_default_value
-                           when repeated? then ::Protobuf::Field::FieldArray.new(self).freeze
-                           when required? then nil
+        @default_value ||= if optional? || required?
```
zanker: Isn't adding `required?` unnecessary because `required?` should always have a value.

lib/protobuf/field/base_field.rb
zanker: :+1:

lib/protobuf/field/bytes_field.rb
```diff
@@ -50,33 +49,17 @@ def wire_type
...
-              @values.delete(field.name)
-            when ::Protobuf::Message
-              @values[field.name] = val.dup
-            else
-              fail TypeError, "Unacceptable value #{val} for field #{field.name} of type #{field.type_class}"
-            end
-          end
+      def coerce!(value)
+        case value
+        when String, Symbol
```
zanker: Why coerce a string to a string, are you trying to dup it?

zanker: Actually apparently calling `to_s` on a string doesn't do anything so whatever!

lib/protobuf/field/enum_field.rb
```diff
@@ -32,30 +32,18 @@ def enum?
         true
       end
 
+      def coerce!(value)
+        enum_value = type_class.fetch(value)
+        fail TypeError, "Invalid Enum value: #{value.inspect} for #{name}" unless enum_value
```
zanker: Is `type_class` a hash and it's calling `fetch` on the hash, or is it a custom class? If it's the standard implementation of `Hash#fetch`, this wouldn't ever do something unless somehow an enum value can be nil.

zachmargolis: No this is an enum, and fetch is the API to get th enum object from the tag or bame

lib/protobuf/message.rb
zanker: If you're going to keep `compact!`, you should remove the `dup` and just do `value = value.compact`. You can drop an array iteration that way and get the same behavior.

lib/protobuf/message.rb
```diff
@@ -140,20 +140,54 @@ def to_proto
...
+          else
+            fail TypeError, <<-TYPE_ERROR
+                Expected repeated value of type '#{field.type_class}'
+                Got '#{value.class}' for repeated protobuf field #{field.name}
+            TYPE_ERROR
+          end
+
+          if value.nil? || (value.respond_to?(:empty?) && value.empty?)
```
zanker: This `respond_to?(:empty)` is unnecessary it looks like? You already check if it's an `Array` and if not error. You could remove the `nil?` too because a `NilClass` would raise an error.

lib/protobuf/message.rb
```diff
@@ -140,20 +140,54 @@ def to_proto
...
+        end
+      else
+        fail ArgumentError, "invalid field name=#{name.inspect}"
       end
     end
 
     def []=(name, value)
       if (field = self.class.get_field(name, true))
-        __send__(field.setter, value) unless value.nil?
+        @encode = nil
```
zanker: I don't see this instance variable used anywhere

lib/protobuf/message/fields.rb
```diff
@@ -92,20 +94,42 @@ def get_field(name_or_tag, allow_extension = false)
...
+        def extension_shortcuts
+          @extension_shortcuts ||= Set.new
+        end
+
         def define_field(rule, type_class, field_name, tag, options)
           raise_if_tag_collision(tag, field_name)
           raise_if_name_collision(field_name)
 
-          field = ::Protobuf::Field.build(self, rule, type_class, field_name, tag, options)
+          if options[:extension]
```
zanker: Perhaps worth typing out a sample string of what is what throughout this? Little confusing to read.

lib/protobuf/message/fields.rb
```diff
@@ -92,20 +94,42 @@ def get_field(name_or_tag, allow_extension = false)
...
           field_store[field_name.to_s] = field
           # defining a new field for the message will cause cached @all_fields, @extension_fields,
           # and @fields to be incorrect; reset them
           @all_fields = @extension_fields = @fields = nil
+        end
 
-          define_method("#{field_name}!") do
-            @values[field_name]
+        def remove_existing_accessors(accessor)
+          ["", "=", "!", "?"].each do |modifier|
```
zanker: Should make this a constant imo

lib/protobuf/message/fields.rb
```diff
@@ -92,20 +94,42 @@ def get_field(name_or_tag, allow_extension = false)
...
+            method = "#{accessor}#{modifier}".to_sym
+            remove_method(method) if instance_methods(false).include?(method)
```
zanker: Is `remove_method` a method? I can't seem to hit it in Ruby.

lib/protobuf/field/enum_field.rb
zanker: Ahhh ok

lib/protobuf/message.rb
nerdrew: Good call. Note though: We copied the code over with minimal changes. No objection to cleaning as we go though.

lib/protobuf/field/base_field.rb
nerdrew: Do we need to? `foo?` will call `foo`, :boom: => deprecation warning, no?

lib/protobuf/field/base_field.rb
nerdrew: lol, ya, but protoc allows default values on required fields, sooooo... we should handle it?

lib/protobuf/field/bytes_field.rb
nerdrew: Hmmm, should we `to_s.dup`?

lib/protobuf/message/fields.rb
zachmargolis: yes, [Module#remove_method](http://ruby-doc.org/core-1.9.3/Module.html#method-i-remove_method)

lib/protobuf/message.rb
nerdrew: ya, but it is used in https://github.com/ruby-protobuf/protobuf/blob/master/lib/protobuf/message/serialization.rb#L51 which is included.

lib/protobuf/field/base_field.rb
embark: Could do it for clarity right?

embark:
re: https://github.com/ruby-protobuf/protobuf/pull/302#issuecomment-183152422

We don't yet actually use fully qualified names, but there's a PR incoming where we will add a test for it

lib/protobuf/message/fields.rb
zanker: Weird, not sure why I couldn't access it. What happens if you try and call `remove_method` on something that doesn't exist? It would be better to not have O(n*4) + whatever the performance of doing `instance_methods(false)` is, if you can instead do `remove_method(method)`, or alternatively call `rescue nil` if we think it mostly exists.

lib/protobuf/message.rb
zanker: Ah SGs

lib/protobuf/field/bytes_field.rb
zanker: If you want to ensure you clone, then I would split it into two cases, one is `String` where you call `dup` and the other is `Symbol` where you call `to_s`. Better to reduce the # of method calls wherever we can.

lib/protobuf/field/base_field.rb
zanker: lol welp

lib/protobuf/field/base_field.rb
```diff
@@ -69,10 +69,12 @@ def default
...
+                             typed_default_value
+                           elsif repeated?
+                             ::Protobuf::Field::FieldArray.new(self).freeze
+                           else
+                             fail "something went very wrong"
```
zachmargolis: maybe something like "unable to produce default value" or something? very ominous message

embark: Well this code should be unreachable actually, so it should be ominous -- it means a field isn't optional, required, or repeated. We could make change the message though to still be ominous while more descriptive :)

lib/protobuf/field/bytes_field.rb
embark: Done

embark:
All comments are addressed, interested in review by someone with write access to the repo :)

embark:
Hey all, I'm looking for reviews from some people with write access because there will be a few commits following up to this with the goal of adding custom option support to the gem. Pinging a couple people who have merged recently (many apologies for the noise).

CC @film42 @liveh2o

lib/protobuf/field/base_field.rb
```diff
@@ -122,10 +124,10 @@ def required?
 
       # FIXME: need to cleanup (rename) this warthog of a method.
       def set(message_instance, bytes)
-        return message_instance.__send__(setter, decode(bytes)) unless repeated?
-        return message_instance.__send__(getter) << decode(bytes) unless packed?
+        return message_instance[name] = decode(bytes) unless repeated?
```
liveh2o: Does this still flow through the setter? I'm exactly sure why we needed to be calling the setter here, but I want to make sure this isn't unintentionally altering behavior.

liveh2o: Though perhaps that is the point here?

embark: The `[]=` method now is the setter and `[]` is now the getter, so it should be the same. The old setter and getter still exist for runtime access of the proto, but they alias to the former two methods now.

For just a little more background, `[]` and `[]=` used to alias to the setter and getter instead, but we switched that around so that we didn't have this weird dynamic usage of the setters and getters.

liveh2o: Got it. No more `__send__(xxx)`.

brianstien:
@embark I don't have write access, but for any reviewers that do, I reviewed this and :+1:.

Flipping around the `[]`, `[]=` with the setters makes the internals easier to read.  I also like using the `coerce!` interface instead of overriding the `define_setter` method.

lib/protobuf/message.rb
film42: We've undone this `@encode` memoization because we found that we can't bust the cash via appending to a repeated getter without some super hacks. We can remove this.

film42:
Just remove the `@encode` and I'll cut a release. Thank you for your contribution! :)

embark:
Many thanks to all the reviews and responses! -- Removed `@encode` so all should be well

embark:
CC @film42 all green!

film42:
:shipit:

film42:
Hmm. I used to have push access to rubygems.. looks like someone is going to need to add me again. Tomorrow I'll cut this as `3.7.0.pre0`.

embark:
@film42 Thanks! There will be a few PRs to come to get support for custom options if you want to wait for that before cutting a new release, though it's up to you.

embark:
FYI @film42 @liveh2o @zachmargolis @nerdrew I just discovered that this changed the behavior of the `[]=` setter to be (arguably) correct in the case when `nil` is passed to it as a value.

Old behavior for repeated fields:
```
% pry(main)> message = ::Test::TestMessage.new(repeated: [1,2])
=> #<Test::TestMessage repeated =[1, 2]>

% pry(main)> message['repeated'] = nil
=> nil

% pry(main)> message
=> #<Test::TestMessage repeated =[1, 2]>
```
note that nothing happens when `nil` is passed in.

New behavior for repeated fields:
```
% pry(main)> message['repeated'] = ::Test::TestMessage.new(repeated: [1,2])
=> #<Test::TestMessage repeated =[1, 2]>

% pry(main)> message['repeated'] = nil
TypeError:                 Expected repeated value of type 'Protobuf::Field::Int32Field'
                Got 'NilClass' for repeated protobuf field rep

% pry(main)> message
=> #<Test::TestMessage rep=[1, 2]>
```

Old behavior for optional fields:
```
% pry(main)> message = ::Test::TestMessage.new(optional: 1)
=> #<Test::TestMessage optional = 1>

% pry(main)> message['optional'] = nil
=> nil

% pry(main)> message
=> #<Test::TestMessage optional = 1>
```
again, nothing happens to the field, it isn't even cleared.

New behavior for optional fields:
```
% pry(main)> message = ::Test::TestMessage.new(optional: 1)
=> #<Test::TestMessage optional = 1>

% pry(main)> message['optional'] = nil
=> nil

% pry(main)> message
=> #<Test::TestMessage optional = 0>
```
now the optional field has been cleared.

**NOTE**: *the new behavior now matches the behavior that the dynamically defined setters `message.repeated` and `message.optional` has always had.* Arguably it was a bug to silently do nothing in the `[]=` setter when `nil` was passed in.

Issue #196, opened 2014-06-05
Trouble when extensions and regular fields have naming conflicts


jkingdon:
In protobufs, an extension is allowed to have the same name as a field, and we can't currently handle it. Solutions might include a syntax like `buf.extension[:extname]` similar to other protobuf libraries (see https://developers.google.com/protocol-buffers/docs/proto#extensions where it talks about SetExtension), a way of establishing a mapping between the different names, or something else.

nerdrew:
It seems this was never resolved. Is the plan outlined above still valid?

We have been working on support for custom options and need a solution for duplicate extension names. If the above plan is still valid, we'll implement it.

cc @embark

nerdrew:
Test case:

boom.proto
```
syntax = "proto2";

message Boom {
  optional uint64 foo = 1;
  extensions 2 to 10;
}

message Bam {
  extend Boom {
    optional int32 foo = 2;
  }
}
```

bad.pb.rb
```
require 'protobuf/message'

class Boom < ::Protobuf::Message; end
class Bam < ::Protobuf::Message; end

class Boom
  optional :uint64, :foo, 1
  # Extension Fields
  extensions 2...11
end

class ::Boom < ::Protobuf::Message
  optional :int32, :foo, 2, :extension => true
end
```

good.pb.rb
```
require 'protobuf/message'

class Boom < ::Protobuf::Message; end
class Bam < ::Protobuf::Message; end

class Boom
  optional :uint64, :foo, 1
  # Extension Fields
  extensions 2...11
end

class ::Boom < ::Protobuf::Message
  optional :int32, :".bam.foo", 2, :extension => true
end
```

using good.pb.rb
```
require "good.pb"

boom = Boom.new(goat: 10)
boom.goat = 12
boom.extensions[:".foo.goat"] = 10
boom.extensions[:".bam.goat"] = 5
```

Question re: https://github.com/ruby-protobuf/protobuf/issues/196#issuecomment-52879975 point 1
The first extension field will define an un-namespaced getter/setter if one does not exist. What will happen with a subsequent extension field? Should it remove (or redefine to be an error) the un-namespaced getter/setter (assuming it was created for an extension field)?

PR #303, opened 2016-02-17
Namespace extension fields

nerdrew:
This PR is branched off https://github.com/ruby-protobuf/protobuf/pull/302, so the first commit can be ignored.

cc @embark

### DIFF
--- a/lib/protobuf/generators/field_generator.rb
+++ b/lib/protobuf/generators/field_generator.rb
@@ -20,6 +20,12 @@ module Protobuf
       attr_reader :field_options
 
       def applicable_options
+        # Note on the strange use of `#inspect`:
+        #   :boom.inspect #=> ":boom"
+        #   :".boom.foo".inspect #=> ":\".boom.foo\""
+        # An alternative to `#inspect` would be always adding double quotes,
+        # but the generatated code looks un-idiomatic:
+        #   ":\"#{:boom}\"" #=> ":\"boom\"" <-- Note the unnecessary double quotes
         @applicable_options ||= field_options.map { |k, v| "#{k.inspect} => #{v}" }
       end
 
@@ -64,11 +70,7 @@ module Protobuf
       end
 
       def name
-        @name ||= if descriptor.name.include?('.')
-                    ":#{descriptor.name.inspect}"
-                  else
-                    ":#{descriptor.name}"
-                  end
+        @name ||= descriptor.name.to_sym.inspect
       end
 
       def number
--- a/lib/protobuf/generators/field_generator.rb
+++ b/lib/protobuf/generators/field_generator.rb
@@ -20,7 +20,7 @@ module Protobuf
       attr_reader :field_options
 
       def applicable_options
-        @applicable_options ||= field_options.map { |k, v| ":#{k} => #{v}" }
+        @applicable_options ||= field_options.map { |k, v| "#{k.inspect} => #{v}" }
       end
 
       def default_value
@@ -64,7 +64,11 @@ module Protobuf
       end
 
       def name
-        @name ||= ":#{descriptor.name}"
+        @name ||= if descriptor.name.include?('.')
+                    ":#{descriptor.name.inspect}"
+                  else
+                    ":#{descriptor.name}"
+                  end
       end
 
       def number
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -21,7 +21,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 493
 # Configuration parameters: AllowURI, URISchemes.
 Metrics/LineLength:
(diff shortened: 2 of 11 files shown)


lib/protobuf/generators/field_generator.rb
```diff
@@ -20,7 +20,7 @@ class FieldGenerator < Base
       attr_reader :field_options
 
       def applicable_options
-        @applicable_options ||= field_options.map { |k, v| ":#{k} => #{v}" }
```
zanker: I'm confused by what this is for, same with the one below

zanker:
Seems fine, confused by `inspect` though

lib/protobuf/generators/field_generator.rb
zachmargolis: it's for extension names with a dot in the middle -- wouldn't be valid ruby syntax with `:foo.bar` so it has to be `:"foo.bar"`

embark: You can't do `:.string` without ruby complaining, you have to do `:".string"`. Now that we are namespacing options and extensions in general, they will start with `.`

embark: Dang it @zachmargolis beat me while I had typed out my answer but failed to press send for 20 minutes :P

zanker: My suggestion would be to explicitly use double quotes rather than relying on inspect since it's not totally obvious. Or maybe a comment to clarify?

zachmargolis:
So excited for this! Should we do a bigger version bump than a patch since it may be not bavvkwards compat?

embark:
Agreed that it is not currently backwards compatible in this case:

Before:
```
> ::Example::Message.new[:extension_field]
=> false
```

After:
```
> ::Example::Message.new[:extension_field]
=> ArgumentError: invalid field name=:extension_field
> ::Example::Message.new[:".package.Namespaced.extension_field"]
=> false
```

lib/protobuf/generators/field_generator.rb
nerdrew: So:

```
@applicable_options ||= field_options.map { |k, v| ":\"#{k}\" => #{v}" }
``` ?

Examples:
```
[1] pry(main)> sym = ".boom.foo".to_sym
=> :".boom.foo"
[2] pry(main)> puts sym.inspect
:".boom.foo"
=> nil
[3] pry(main)> puts %[:"#{sym}"]
:".boom.foo"
=> nil
```

**Comment by nerdrew** [comment]
@embark True. It is not backward compatible. To make it backward compatible, we can store the unnamespaced extension name in the message field store (as both a string and a symbol), but it is getting a little crazy in there.

The `field_store` would then have the following keys that all refer to the name field: `tag`, `field_name` (a symbol), `field_name.to_s`, unnamespaced `field_name`, and unnamespaced `field_name.to_s`.

```
field_store[1] == field_store[:extension_field] == field_store["extension_field"] == field_store[:".my_package.extension_field"] == field_store[".my_package.extension_field"]
```

No judgements, but, hmmmm.... yes judgements.

embark:
@zachmargolis or @film42 thoughts on keeping it backwards compatible now with deprecated accessors, or just doing a bigger version bump? It will be so nice to support extensions with the same name in for the same message.

zachmargolis:
@embark I think that as good software practices, it would be good make a small bump with a deprecation warning and then a big bump with the breaking change? That is sort of what was outlined in in #196: https://github.com/ruby-protobuf/protobuf/issues/196#issuecomment-52879975

liveh2o:
It might be ugly, but a clean upgrade path is how we typically try to roll. We might be approaching time for a major version bump where we can cleanup a lot of the "cruft" that we've accumulated.

lib/protobuf/generators/field_generator.rb
nerdrew: Addressed with a comment.

(Not applicable to proto field names, but inspect is nice: `:'boom"'.inspect #=> ":\"boom\\\"\""`)

embark:
Hey guys so we updated the branch to be backwards compatible. The specs of the new commit show the new accessors that we allow, basically:
`field_store[1] == field_store[:extension_field] == field_store["extension_field"] == field_store[:".my_package.extension_field"] == field_store[".my_package.extension_field"]` works

zachmargolis:
yayyyyyy LGTM :+1: :sunglasses: :tada:

liveh2o:
One last look, @film42 or @mmmries?

embark:
@film42 @mmmries beginning of the week ping :)

film42:
:shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit:

PR #305, opened 2016-02-18
Remove memoization on encode

### LINKED ISSUE #304: Memoized encoding does not properly handle repeated field operations
The changes introduced in https://github.com/ruby-protobuf/protobuf/pull/293, performing operations on repeated fields is no longer handled properly after the memoized encoding has been computed.

```ruby
a.encode
a.some_repeated_field << "new-value"
a.encode # doesn't contain "new-value"
```

liveh2o:
Changing values in repeated fields does not clear the memoized encoded value. This means that changes to repated fields after the memoized encoding has been computed will never be encoded. The pattern works for other field types because the memoized value is being cleared in the setter on the message before we call into the field. The field itself is not involved in clearing the memoized value at all.

Repeated fields are a different story. If we were completely resetting a repeated field to a new array, the memoized encoding would be correctly cleared, but because the field array we are pushing into doesn't have any concept of a message, let alone the specific message instance the field belongs to, there is no way to clear the memoized encoded message.

We need to remove the memoization until we can find a better approach.

Fixes #304.

// @film42 @abrandoned @goddardc

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -198,7 +198,6 @@ module Protobuf
 
         message_class.class_eval do
           define_method(method_name) do |val|
-            @encode = nil
             if val.is_a?(Array)
               val = val.dup
               val.compact!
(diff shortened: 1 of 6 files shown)


film42:
Restarted the job, not sure why some of these things were failing (looks vm specific). I agree, we're going  to need to attack this in a different way.

liveh2o:
Yeah, the one you restarted failed as well. Ran it a third time and things are now passing! Ready for review.

film42:
:shipit:

PR #308, opened 2016-02-28
Small optimizations

film42:
This just adds a few speedups.
1. `class_eval` string - This turned out to be about 40% faster for setters and 48% faster for getters (for both MRI and Jruby). Here's my benchmark: https://gist.github.com/film42/8efd4810d3c5a5e12291.
2. Varint - This takes the optimized varint out of https://github.com/ruby-protobuf/protobuf/pull/297.
3. `nonzero?` vs `!= 0` - Declarative methods are nice, but simple operators are still very readable and more performant, especially in these hot code paths.

Note: I also added `benchmark-ips` as a dev dependency because I end up using it a lot when working with protobuf.

cc @abrandoned @zachmargolis @embark

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -177,16 +177,19 @@ module Protobuf
       #
 
       def define_accessor(simple_field_name, fully_qualified_field_name)
-        message_class.class_eval do
-          define_method("#{simple_field_name}!") do
-            @values[fully_qualified_field_name]
+        message_class.class_eval <<-ruby, __FILE__, __LINE__
+          def #{simple_field_name}!
+            @values[:#{fully_qualified_field_name}]
           end
-        end
 
-        message_class.class_eval do
-          define_method(simple_field_name) { self[fully_qualified_field_name] }
-          define_method("#{simple_field_name}=") { |v| self[fully_qualified_field_name] = v }
-        end
+          def #{simple_field_name}
+            self[:#{fully_qualified_field_name}]
+          end
+
+          def #{simple_field_name}=(value)
+            self[:#{fully_qualified_field_name}] = value
+          end
+        ruby
 
         return unless deprecated?
 
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -179,15 +179,15 @@ module Protobuf
       def define_accessor(simple_field_name, fully_qualified_field_name)
         message_class.class_eval <<-ruby, __FILE__, __LINE__
           def #{simple_field_name}!
-            @values[:#{fully_qualified_field_name}]
+            @values[:"#{fully_qualified_field_name}"]
           end
 
           def #{simple_field_name}
-            self[:#{fully_qualified_field_name}]
+            self[:"#{fully_qualified_field_name}"]
           end
 
           def #{simple_field_name}=(value)
-            self[:#{fully_qualified_field_name}] = value
+            self[:"#{fully_qualified_field_name}"] = value
           end
         ruby
 
--- a/lib/protobuf/varint_pure.rb
+++ b/lib/protobuf/varint_pure.rb
@@ -1,12 +1,12 @@
 module Protobuf
   module VarintPure
     def decode(stream)
-      value = index = 0
+      value = shift = 0
       begin
         byte = stream.readbyte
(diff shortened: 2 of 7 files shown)


lib/protobuf/varint_pure.rb
```diff
@@ -1,12 +1,12 @@
 module Protobuf
   module VarintPure
     def decode(stream)
-      value = index = 0
+      value = shift = 0
       begin
         byte = stream.readbyte
-        value |= (byte & 0x7f) << (7 * index)
```
embark: Curious, is the hex actually slower than the dec?

embark:
:+1: seems good, though I prefer the hex unless it's slower :)

lib/protobuf/varint_pure.rb
film42: They're about the same, but hex is slower:

```
irb(main):002:0> Benchmark.ips do |x|
irb(main):003:1* x.report("hex") { 3 + 0x7F }
irb(main):004:1> x.report("dec") { 3 + 127 }
irb(main):005:1> end

Calculating -------------------------------------
                 hex   130.033k i/100ms
                 dec   119.055k i/100ms
-------------------------------------------------
                 hex      8.761M (± 4.6%) i/s -     43.691M
                 dec      9.034M (± 5.7%) i/s -     45.003M
```

If you think it reads better as a hex (fine either way for me; google uses both), I'll change it back.

embark: Seems pretty variable:
```
Warming up --------------------------------------
                 hex   186.594k i/100ms
                 dec   195.158k i/100ms
Calculating -------------------------------------
                 hex      9.104M (± 6.8%) i/s -     45.342M
                 dec      8.828M (± 6.4%) i/s -     44.106M
```

But I don't want to nitpick, either way is fine :). Hex is nicer for visualizing bitwise arithmetic, but I'm was mostly just curious :)

zachmargolis:
Seems legit, LGTM :+1:

mmmries:
:+1: this looks good to me. No strong preference on `hex` vs `dec`.

lib/protobuf/field/base_field.rb
```diff
@@ -172,16 +172,19 @@ def wire_type
...
-        message_class.class_eval do
-          define_method(simple_field_name) { self[fully_qualified_field_name] }
-          define_method("#{simple_field_name}=") { |v| self[fully_qualified_field_name] = v }
-        end
+          def #{simple_field_name}
+            self[:#{fully_qualified_field_name}]
+          end
+
+          def #{simple_field_name}=(value)
+            self[:#{fully_qualified_field_name}] = value
```
ryanbjones: do we know that `fully_qualified_field_name` will be a string and not a symbol?

film42: Symbol or string, it should be the same: `puts :ok #=> "ok"`

ryanbjones: :+1:

film42: Actually, maybe it should look like `:"#{fully_qualified_field_name}"` to support weird symbols like `:"some.sym.yolo"`.

ryanbjones: spec to verify?

film42: Updated. Now that the full qualified field name code is in master, I rebased to see how the current implementation would work and got errors because `:some.thing` is not valid ruby. Change this to `:"some.thing"` works just fine.

lib/protobuf/field/base_field.rb
```diff
@@ -177,16 +177,19 @@ def fully_qualified_name_only!
       #
 
       def define_accessor(simple_field_name, fully_qualified_field_name)
-        message_class.class_eval do
-          define_method("#{simple_field_name}!") do
-            @values[fully_qualified_field_name]
+        message_class.class_eval <<-ruby, __FILE__, __LINE__
+          def #{simple_field_name}!
+            @values[:"#{fully_qualified_field_name}"]
```
film42: I could also do `@values[#{fully_qualified_field_name.inspect}]`, but I don't really have a preference.

abrandoned:
@film42 were we going to pull this in? looks like it needs to be updated if so

film42:
Still planning on it, but working through getting 3.7.0 features into master first, then coming back and doing a full profile again.

PR #309, opened 2016-03-01
Messages fields should be able to be set with to_proto

embark:
A prior commit (#302) introduced a bug so that you would get `Failure/Error: fail TypeError, "Unacceptable value #{value} for field #{field.name} of type #{field.type_class}"` when trying to set a message field with a class that defined a #to_proto method.

CC @zachmargolis @nerdrew @film42

### DIFF
--- a/lib/protobuf/field/message_field.rb
+++ b/lib/protobuf/field/message_field.rb
@@ -9,7 +9,7 @@ module Protobuf
       #
 
       def acceptable?(val)
-        val.is_a?(type_class) || val.respond_to?(:to_hash)
+        val.is_a?(type_class) || val.respond_to?(:to_hash) || val.respond_to?(:to_proto)
       end
 
       def decode(bytes)
@@ -31,17 +31,19 @@ module Protobuf
       end
 
       def coerce!(value)
-        if value.nil?
-          nil
-        elsif value.is_a?(type_class)
-          value
-        elsif value.respond_to?(:to_proto)
-          value.to_proto
-        elsif value.respond_to?(:to_hash)
-          type_class.new(value.to_hash)
-        else
-          fail TypeError, "Expected value of type '#{type_class}' for field #{name}, but got '#{value.class}'"
-        end
+        return nil if value.nil?
+
+        coerced_value = if value.respond_to?(:to_proto)
+                          value.to_proto
+                        elsif value.respond_to?(:to_hash)
+                          type_class.new(value.to_hash)
+                        else
+                          value
+                        end
+
+        return coerced_value if coerced_value.is_a?(type_class)
+
+        fail TypeError, "Expected value of type '#{type_class}' for field #{name}, but got '#{value.class}'"
       end
 
     end
--- a/spec/lib/protobuf/field/message_field_spec.rb
+++ b/spec/lib/protobuf/field/message_field_spec.rb
@@ -1,0 +1,79 @@
+require 'spec_helper'
+
+RSpec.describe Protobuf::Field::MessageField do
+  let(:field_message) do
+    Class.new(::Protobuf::Message) do
+      optional :int32, :field, 1
+    end
+  end
+
+  let(:message) do
+    Class.new(::Protobuf::Message) do
+      optional FieldMessage, :message_field, 1
+    end
+  end
+
+  before do
+    stub_const('FieldMessage', field_message)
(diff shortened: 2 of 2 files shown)


spec/lib/protobuf/field/message_field_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  class ToProtoMessageField
+    def to_proto
+      MessageField.new(:field => 42)
+    end
+  end
+
+  class ToProtoIsForWrongMessage
+    def to_proto
+      SomeMessage.new()
```
zachmargolis: style nit, drop the `()` when there are no arguments

embark: fixed by running rubocop :)

spec/lib/protobuf/field/message_field_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'spec_helper'
+
+RSpec.describe Protobuf::Field::MessageField do
+  class MessageField < ::Protobuf::Message
```
zachmargolis: just as a heads up, this will leak the name `MessageField` into the global namespace in all specs, see `Class.new` in things like `enum_spec.rb` and `message_spec.rb` for how to define anonymous classes in tests that don't leak?

it's not the end of the world but a good practice IMO in case some other test tries to make a class named `MessageField`

zachmargolis:
LGTM! thanks for fixing

spec/lib/protobuf/field/message_field_spec.rb
embark: Ah right, thanks! FIxed

nerdrew:
+1

embark:
@liveh2o @film42 Sorry for the ping, but does this look okay? Don't like to leave bugs around.

lib/protobuf/field/message_field.rb
```diff
@@ -35,8 +35,8 @@ def coerce!(value)
           nil
         elsif value.is_a?(type_class)
           value
-        elsif value.respond_to?(:to_proto)
-          value.to_proto
+        elsif value.respond_to?(:to_proto) && (proto_value = value.to_proto).is_a?(type_class)
```
liveh2o: Is this check necessary? If so, the complexity of a compound conditional that also sets a variable that is used as the return value is something we try to eliminate. If the goal is to avoid having to write the failure line twice, then I'd rewrite the method:

```ruby
def coerce!(value)
  return nil if value.nil?

  coerced_value = if value.respond_to?(:to_proto)
                    value.to_proto
                  elsif value.respond_to?(:to_hash)
                    type_class.new(value.to_hash)
                  else
                    value
                  end
  
  return coerced_value if coerced_value.is_a?(type_class)

  fail TypeError, "Expected value of type '#{type_class}' for field #{name}, but got '#{value.class}'"
end
```

To me, this is much clearer. It basically has three steps:

1. Ignore nil values
2. Try and coerce the value
3. Fail unless the coerced value is the correct type

What do you think?

lib/protobuf/field/message_field.rb
```diff
@@ -9,7 +9,7 @@ class MessageField < BaseField
       #
 
       def acceptable?(val)
-        val.is_a?(type_class) || val.respond_to?(:to_hash)
+        val.is_a?(type_class) || val.respond_to?(:to_hash) || val.respond_to?(:to_proto)
```
liveh2o: Definitely in favor of this!

bacon

liveh2o:
Don't hesitate to ping if you're not getting a response. We need to stay on top of this stuff.

lib/protobuf/field/message_field.rb
zachmargolis: @liveh2o I think this patch is pretty reasonable as-is, I'm happy leaving it as @embark proposed?

embark: I'm actually good with the proposal, it might be more readable and sounds like it's faster? I changed it as suggested.

embark:
Comments addressed :)

liveh2o:
:shipit:

liveh2o:
It looks like we've got a few PRs stacking up. I'll check with @film42 and see we can cut a release tomorrow.

PR #310, opened 2016-03-02
Empty values should not clear a protobuf field

embark:
Before commit #302, empty? values never cleared a proto field unless the field was a Fixnum or boolean type (which didn't make any sense, and was probably unused, anyway).

### DIFF
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -178,7 +178,7 @@ module Protobuf
             @values[field.name].replace(value)
           end
         else
-          if value.nil? || (value.respond_to?(:empty?) && value.empty?)
+          if value.nil?
             @values.delete(field.name)
           elsif field.acceptable?(value)
             @values[field.name] = field.coerce!(value)
--- a/spec/lib/protobuf/message_spec.rb
+++ b/spec/lib/protobuf/message_spec.rb
@@ -181,64 +181,6 @@ RSpec.describe Protobuf::Message do
       test_enum = Test::EnumTestMessage.new { |p| p.non_default_enum = 2 }
       expect(test_enum.non_default_enum).to eq(2)
     end
-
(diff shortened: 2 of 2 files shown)


embark:
CC @nerdrew @zachmargolis @film42

zachmargolis:
Can we get some documentation on the expected behavior with specs, please?

embark:
Yeah, but I'm going to wait for #309 to be merged actually because I want to use the spec file created there

liveh2o:
So this basically makes it so you can set a field to an empty string or a repeated field to an empty array? Is there any other behavior this might alter?

embark:
@liveh2o Technically this might make it so you can no longer pass any type that responds to `empty?` to the Fixnum/Boolean fields. For example, `message.int_field = ""` would have cleared the integter field. 

This is because we would allow you to clear the Fixnum/Boolean fields with an `empty?` value, _without_ doing the `acceptable?` check first. But that is very weird behavior.

liveh2o:
Yeah, I think that makes sense. Any objections, @film42 or @mmmries?

embark:
@film42 I'd like to get this and #312 in at least before cutting a new release, because they fix some bugs introduced by my recent changes

film42:
I just merged #309 , so if you want to get some specs in place, I'll hold off until this and #312 get merged.

embark:
@film42 @zachmargolis added tests, so this should be good to go

zachmargolis:
LGTM!

film42:
Specs failing?

embark:
Whoops, I probably just fixed it. Accidentally pushed some things to this branch earlier and forgot to push the revert

film42:
:shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit: :tada:

PR #312, opened 2016-03-17
Preserve raw RPC name (since #underscore can be lossy)

zachmargolis:
cc @nerdrew @embark 

We found some RPC names that were not the same in a roundtrip of `name.underscore.camelize` so this preserves the raw data

### DIFF
--- a/lib/protobuf/generators/service_generator.rb
+++ b/lib/protobuf/generators/service_generator.rb
@@ -15,10 +15,10 @@ module Protobuf
       end
 
       def build_method(method_descriptor)
-        name = method_descriptor.name
         request_klass = modulize(method_descriptor.input_type)
         response_klass = modulize(method_descriptor.output_type)
-        "rpc :#{name.underscore}, #{request_klass}, #{response_klass}"
+        name = ENV.key?('PB_USE_RAW_RPC_NAMES') ? method_descriptor.name : method_descriptor.name.underscore
+        "rpc :#{name}, #{request_klass}, #{response_klass}"
       end
 
     end
--- a/spec/lib/protobuf/generators/service_generator_spec.rb
+++ b/spec/lib/protobuf/generators/service_generator_spec.rb
@@ -41,6 +41,15 @@ end
     it 'returns a string identifying the given method descriptor' do
       expect(subject.build_method(service.method.first)).to eq("rpc :search, FooRequest, FooResponse")
     end
+
+    context 'with PB_USE_RAW_RPC_NAMES in the environemnt' do
+      before { allow(ENV).to receive(:key?).with('PB_USE_RAW_RPC_NAMES').and_return(true) }
+
+      it 'uses the raw RPC name and does not underscore it' do
+        expect(subject.build_method(service.method.first)).to eq("rpc :Search, FooRequest, FooResponse")
+        expect(subject.build_method(service.method.last)).to eq("rpc :FooBar, ::Foo::Request, ::Bar::Response")
+      end
+    end
   end
 
 end


zachmargolis:
Another option would be to add a compiler env var flag to just use the raw name instead, would simplify a lot of the code

Ex: `PB_RAW_RPC_NAMES=true` or something

film42:
Looks good to me. I'm looking to get some of these bigger PRs merged first, but I'll grab this after.

lib/protobuf/rpc/service.rb
```diff
@@ -103,8 +103,9 @@ class << self
       # This methods is only used by the generated service definitions
       # and not useful for user code.
       #
-      def self.rpc(method, request_type, response_type)
-        rpcs[method] = RpcMethod.new(method, request_type, response_type)
+      def self.rpc(method, request_type, response_type, options = {})
+        name = options[:name] || method
```
embark: Why not just always use the exact name? It's not a breaking change

zachmargolis: Because of old code that goes to look up `service.rpcs[old_name]` will fail with

embark: Oh `name` was already used, I see. You could also make a new variable

embark:
@zachmargolis this code is also probably broken if the rpc names collide when underscored, right? Like in this issue: https://github.com/grpc/grpc/issues/5812

zachmargolis:
@embark I'll redo this with the env var flag outlined above -- maybe we could make the "raw" name the default behavior in the near future

lib/protobuf/generators/service_generator.rb
```diff
@@ -15,10 +15,10 @@ def compile
       end
 
       def build_method(method_descriptor)
-        name = method_descriptor.name
         request_klass = modulize(method_descriptor.input_type)
         response_klass = modulize(method_descriptor.output_type)
-        "rpc :#{name.underscore}, #{request_klass}, #{response_klass}"
+        name = ENV.key?('PB_RAW_RPC_NAMES') ? method_descriptor.name : method_descriptor.name.underscore
```
film42: What about `PB_PRESERVE_RPC_NAMES` or `PB_USE_RAW_RPC_NAMES`? Our env names tend to have an action associated with them.

film42:
Added one comment about naming, but otherwise this is :+1: to me. Let's get this fixed/ merged and then release a pre. I'm not sure if it should be `3.7.0.pre0` or `4.0.0.pre0` but it's much too late to debate about that for tonight. If you have an opinion one way or the other, let me know :).

lib/protobuf/generators/service_generator.rb
liveh2o: I don't have a particular preference, but :+1: for including an action or verb.

embark:
Thanks, looks good!

zachmargolis:
Done! Used `PB_USE_RAW_RPC_NAMES`

film42:
Grrr. Looks like rubo:cop: has a new release out.

zachmargolis:
How did Rubocop change? Looks like the version was pinned in the gemspec back in 86ef9075a2115904b59612c6152717c475164ea9

film42:
Looks like the ruby parser gem had a breaking changed released as a patch verison https://github.com/bbatsov/rubocop/pull/2984#issuecomment-201334316. I'm not able to lock it down and fix it though, so I'm honestly not sure what the issue is. I'll let them sort it out.

:shipit:

film42:
Thank you for your PR, btw :)

film42:
`3.7.0.pre0` has been released.

PR #313, opened 2016-03-21
Field values should be stored via their fully qualified names

embark:
This fixes a bug where ::Protobuf::Message#{simple_field_name}! would return nil rather than the value of the field, if the field was an extension.
This also fixes a bug where ::Protobuf::Message.field?(name) would always return false if name was the fully qualified name of an extension.

CC @nerdrew @film42 @liveh2o @zachmargolis

### DIFF
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -84,7 +84,7 @@ module Protobuf
 
     def each_field_for_serialization
       self.class.all_fields.each do |field|
-        value = @values[field.getter]
+        value = @values[field.fully_qualified_name]
         if value.nil?
           fail ::Protobuf::SerializationError, "Required field #{self.class.name}##{field.name} does not have a value." if field.required?
           next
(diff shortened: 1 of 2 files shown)


embark:
**IMPORTANT NOTE**: This as-is currently changes the behavior of `::Protobuf::Message#to_hash`:

Before:
```
pry(main)> ::Test::Resource.new(:ext_is_searchable => true).to_hash
=> {:ext_is_searchable=>true}
```

After:
```
pry(main)> ::Test::Resource.new(:ext_is_searchable => true).to_hash
=> {:".test.Searchable.ext_is_searchable"=>true}
```

Any ideas on what the best way to get around this is, if that is desired? We could make `to_hash` be passed an optional parameter, or have a different method entirely that uses fully qualified names in the hash.

embark:
Also I expect tests to fail until #310 is merged

zachmargolis:
LGTM!

film42:
Looks like tests are failing.

embark:
@film42 Yes, tests will fail until  #310 is merged because of that bug

film42:
Sorry! Early morning code reviews be like.. :grin: .

embark:
No worries, thanks for the review :)

film42:
Just merged #310.

film42:
:shipit: :tada: :shipit: :tada: :shipit:

film42:
`3.7.0.pre0` has been released.

embark:
Awesome!!

PR #314, opened 2016-03-24
Properly encode and decode negative enum values

embark:
From proto2 guide: Enumerator constants must be in the range of a 32-bit integer. Since enum values use varint encoding on the wire, negative values are inefficient and thus not recommended.

Before this change, enum values were incorrectly treated as uint values, and negative values would cause this error in the Varint gem:
`no implicit conversion from nil to integer`

### DIFF
--- a/lib/protobuf/field/enum_field.rb
+++ b/lib/protobuf/field/enum_field.rb
@@ -1,8 +1,8 @@
-require 'protobuf/field/varint_field'
+require 'protobuf/field/integer_field'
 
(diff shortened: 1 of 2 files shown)


embark:
CC @zachmargolis @nerdrew @film42

nerdrew:
Looks good.

film42:
:shipit: :tada: :shipit: :tada: :shipit: :tada: :shipit:

PR #315, opened 2016-03-28
Add option to name a package from file name if necessary

embark:
Next in the series of PRs that gets us closer to custom options!
We will always need a package name if we want to support custom file options and there isn't a base module to define those custom file options on.

This is how other languages handle an unspecified package name, [I'm pretty sure].

CC @nerdrew @film42


nerdrew:
Looks good to me. Rubocop is complaining, but it is unrelated to this PR.

embark:
Any thoughts @film42 @liveh2o ?

film42:
Sorry for the slow review!! :shipit: !!

PR #316, opened 2016-03-28
Make all things Optionable and fix requires

embark:
In prep for adding support for custom field, enum, message, file, service, and method options

CC @nerdrew @film42 @liveh2o 

Re: requires

`get_option` needs Google::Protobuf::*Options to work (Optionable takes
a Google::Protobuf::*Options class as an argument), so descriptor.pb
needs to get required before using `get_option`.

BUT, to parse descriptor.pb, Protobuf::{Enum,Message,etc} needs to
exist. So, require protobuf/message, then require descriptor.pb

get_option! -> return explicitly set options only

### DIFF
--- a/lib/protobuf/generators/file_generator.rb
+++ b/lib/protobuf/generators/file_generator.rb
@@ -104,7 +104,7 @@ module Protobuf
       end
 
       def print_generic_requires
-        print_require("protobuf/message")
+        print_require("protobuf")
         print_require("protobuf/rpc/service") if descriptor.service.count > 0
         puts
       end
--- a/lib/protobuf/generators/file_generator.rb
+++ b/lib/protobuf/generators/file_generator.rb
@@ -29,6 +29,7 @@ module Protobuf
           print_import_requires
 
           print_package do
+            inject_optionable
             group = GroupGenerator.new(current_indent)
             group.add_enums(descriptor.enum_type, :namespace => [descriptor.package])
             group.add_message_declarations(descriptor.message_type)
@@ -141,6 +142,10 @@ module Protobuf
         token[0] == '.'
       end
 
+      def inject_optionable
+        return if descriptor.package.empty? && !ENV.key?('PB_ALLOW_DEFAULT_PACKAGE_NAME')
+        puts "::Protobuf::Optionable.inject(self) { ::Google::Protobuf::FileOptions }"
+      end
     end
   end
 end
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -21,6 +21,7 @@ module Protobuf
 
     extend ::Protobuf::Message::Fields
     include ::Protobuf::Message::Serialization
+    ::Protobuf::Optionable.inject(self) { ::Google::Protobuf::MessageOptions }
 
     ##
     # Class Methods
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -1,7 +1,3 @@
-require 'protobuf/field'
-require 'protobuf/deprecation'
-require 'protobuf/enum'
-require 'protobuf/exceptions'
 require 'protobuf/message/fields'
 require 'protobuf/message/serialization'
 
--- a/lib/protobuf/optionable.rb
+++ b/lib/protobuf/optionable.rb
@@ -1,17 +1,36 @@
-require 'active_support/concern'
-
 module Protobuf
   module Optionable
-    extend ::ActiveSupport::Concern
-
     module ClassMethods
       def get_option(name)
-        @_optionable_options.try(:[], name)
+        name = name.to_s
+        option = optionable_descriptor_class.get_field(name, true)
+        fail ArgumentError, "invalid option=#{name}" unless option
+        unless option.fully_qualified_name.to_s == name
+          # Eventually we'll deprecate the use of simple names of fields completely, but for now make sure people
+          # are accessing options correctly. We allow simple names in other places for backwards compatibility.
+          fail ArgumentError, "must access option using its fully qualified name: #{option.fully_qualified_name.inspect}"
+        end
+        if @_optionable_options.try(:key?, name)
+          value = @_optionable_options[name]
+        else
+          value = option.default_value
+        end
+        if option.type_class < ::Protobuf::Message
+          option.type_class.new(value)
+        else
+          value
+        end
       end
+
+      def get_option!(name)
+        get_option(name) if @_optionable_options.try(:key?, name.to_s)
+      end
+
+      private
 
       def set_option(name, value = true)
         @_optionable_options ||= {}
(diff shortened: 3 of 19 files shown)


embark:
Tests will fail until #315 is merged

embark:
@zanker @zachmargolis care to review?

lib/protobuf/generators/file_generator.rb
```diff
@@ -138,6 +139,10 @@ def fully_qualified_token?(token)
         token[0] == '.'
       end
 
+      def inject_optionable
+        return if descriptor.package.empty? && !ENV.key?('PB_DEFAULT_PACKAGE_NAME')
+        puts "::Protobuf::Optionable.inject(self) { ::Google::Protobuf::FileOptions }"
```
zanker: Does it auto write STDOUT to file?

lib/protobuf/message.rb
```diff
@@ -238,3 +240,5 @@ def copy_to(object, method)
 
   end
 end
+
+require 'protobuf/descriptors'
```
zanker: Shouldn't this be at the top?

zanker:
At least code wise it looks sane, I don't have a lot of context on the protobuf specifics though.

lib/protobuf/generators/file_generator.rb
embark: Yeah the puts method is intercepted

lib/protobuf/message.rb
embark: No, get_option needs Google::Protobuf::Options to work (Optionable takes
a Google::Protobuf::Options class as an argument), so descriptor.pb
needs to get required before using get_option.

BUT, to parse descriptor.pb, Protobuf::{Enum,Message,etc} needs to
exist. So, require protobuf/message, then require descriptor.pb

I suppose I could add a comment lol

spec/lib/protobuf/optionable_spec.rb
```diff
@@ -1,46 +1,152 @@
...
+    end
+
+    it '.get_option can return an option representing a message' do
+      message_field = ::Protobuf::Field::MessageField.new(
+        OptionableGetOptionTest, :optional, OptionableGetOptionTest, '.package.message_field', 2, 'message_field', {})
+      allow(::Google::Protobuf::MessageOptions).to receive(:get_field).and_return(message_field)
+      expect(OptionableGetOptionTest.get_option(message_field.fully_qualified_name)).to eq(OptionableGetOptionTest.new(:field => 33))
+    end
+
+    it '.get_option! retrieves explicitly an set option' do
```
nerdrew: "an explicitly set option"

nerdrew:
LG. KEEP UP THE GOOD WORK

embark:
Ping @film42 @liveh2o . Tests expected to fail until my other PR is merged, sorry about that

film42:
Just merged #315

lib/protobuf/optionable.rb
```diff
@@ -1,23 +1,59 @@
...
+          # Eventually we'll deprecate the use of simple names of fields completely, but for now make sure people
+          # are accessing options correctly. We allow simple names in other places for backwards compatibility.
+          fail ArgumentError, "must access option using its fully qualified name: #{option.fully_qualified_name.inspect}"
+        end
+        if @_optionable_options.try(:key?, name)
+          value = @_optionable_options[name]
+        else
+          value = option.default_value
+        end
+        if option.type_class < ::Protobuf::Message
```
film42: TIL you can do this in ruby. I think it would be easier on the eyes to read:
```ruby
option.type_class.is_a?(::Protobuf::Message) ? option.type_class.new(value) : value
```

embark: I can switch to a ternary operator, but the `is_a?` behavior won't work on a subclass of `::Protobuf::Message`:

```
class A; end
class B < A;end
B.is_a? A 
=> false
```

I think the `<` operator is the correct method to use here

zachmargolis: Yeah that's because B.class == Class, @embark is right here

film42: Oh! I see. TIL, again. :blush: :+1: .

film42:
Let's get this branch cleaned up and I think it should be good to go.

embark:
What have I done to rubocop :/

embark:
@film42 sorry for the delay, fixed rubocop errors by disabling the broken cop

embark:
@film42 bump :)

zachmargolis:
Hey so I know this is late in the life of this PR but I think we can clean up he markup by a lot, and not have to include that `inject` line if we use the [Class#inherited](http://ruby-doc.org/core-1.9.3/Class.html#method-i-inherited) hook


So instead of having

```
class Message < Protobuf::Message
  Protobuf::Optionable.inject(self)
end
```

We can define `#inherited` on the Message class:

```
class Protobuf::Message
  def self.inherited(subklass)
    Protobuf::Optionable.inject(subklass)
  end
end
```

Which has the same net effect and cleans up the result. WDYT?

embark:
@zachmargolis If I understand what you're saying, I don't think that's going to work the way you want. We're actually injecting **`FileOptions`** on every inner **`module`** that we get from the package name only (none of which extend anything). For example:

```
module Google
  module Protobuf
    ::Protobuf::Optionable.inject(self) { ::Google::Protobuf::FileOptions }
```

The `Protobuf::Message` inject is already inherited by default with this code:
```
module Protobuf
  class Message
    ::Protobuf::Optionable.inject(self) { ::Google::Protobuf::MessageOptions }
```

zachmargolis:
@embark oh you are right I completely misunderstood, thx for explaining

lib/protobuf/optionable.rb
```diff
@@ -1,23 +1,70 @@
...
+    def get_option!(name)
+      self.class.get_option!(name)
+    end
+
+    def self.inject(base_class, extend_class = true, &block)
+      unless block_given?
+        fail ArgumentError, 'missing option class block (e.g: ::Google::Protobuf::MessageOptions)'
+      end
+      if extend_class
+        # Check if optionable_descriptor_class is already defined and short circuit if so.
```
embark: @film42 @zachmargolis FYI just added this logic (plus two tests which cover each case)

embark:
@film42 @nerdrew changed the require logic structure, makes things ultimately more simple I think. Let me know what you think

embark:
ping @film42 @liveh2o for review :)

film42:
Thanks for the ping @embark ! :shipit: 🎉 

Also, thanks for your patience :)

I probably won't be able to cut a pre until tonight, but I'll set a reminder so I don't forget!

embark:
Thanks! Please give it another look before you merge

film42:
Released as `3.7.0.pre1`!

PR #317, opened 2016-03-29
Fix rubocop: lock down the parser version / upgrade rubocop to 0.35

film42:
This should fix rubocop. I locked down the parser version and had to upgrade rubocop in order to make things work again. There was a new cop added, but I like the change so I kept it enabled.

cc @brianstien @embark @mmmries @liveh2o @abrandoned


embark:
LGTM, thanks!!

PR #319, opened 2016-04-13
Revert to old behavior for setting repeated fields to nil

embark:
The direct setter for repeated fields didn't allow nil values, but everything else did (and ignored nils). We accidentally changed that behavior, even though it's awful behavior and should probably be intentionally changed in the future.

CC @nerdrew @film42 @zachmargolis 

Sorry about the breaking changes :/

### DIFF
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -185,7 +185,7 @@ module Protobuf
 
         message_class.class_eval do
           define_method(simple_field_name) { self[fully_qualified_field_name] }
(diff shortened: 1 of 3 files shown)


nerdrew:
Looks good. The old behavior is really wonky.

embark:
@film42 @mmmries any thoughts on this? I know it's wonky, but it's old behavior

film42:
:shipit:

film42:
This will be released with 3.7. Thanks for the patch :)

PR #322, opened 2016-05-12
Allow user defined service directories

quixoten:
Currently, ZMQ assumes that UDP broadcast based service discovery will
fit all needs. This change allows a user to create their own service
directory and plug it into the ZMQ connector.

Our use case will be creating a service directory that is based on a YAML
file generated by puppet. This will allow us to eliminate the need for layer 2
adjacency with consumers of our RPC services, which is becoming a
security concern in our infrastructure.


quixoten:
re-opening against 3-6-stable

PR #324, opened 2016-05-20
Allow user defined service directories

quixoten:
Currently, ZMQ assumes that UDP broadcast based service discovery will
fit all needs. This change allows a user to create their own service
directory and plug it into the ZMQ connector.


film42:
Thank you for merging into `3-6-stable`. I'll merge and cut a release once travis is 💚 .

film42:
These errors have been fixed in master. :shipit:

PR #325, opened 2016-05-29
Track if a repeated field has been deliberately set

embark:
Before, if an unset repeated field was ever read, we would make an empty field array for it. As a consequence, .field? and #{field}! methods would indicate that the field had been set, even though it had only been read from.
This was a weird side effect of reading an unset repeated field and has been corrected.

CC @film42 @nerdrew

Next piece of ruby-protobuf#323

### DIFF
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -94,7 +94,11 @@ module Protobuf
     def field?(name)
       field = self.class.get_field(name, true)
       return false if field.nil?
-      @values.key?(field.fully_qualified_name)
+      if field.repeated?
+        @values.key?(field.fully_qualified_name) && @values[field.fully_qualified_name].present?
+      else
+        @values.key?(field.fully_qualified_name)
+      end
     end
     ::Protobuf.deprecator.define_deprecated_methods(self, :has_field? => :field?)
 
--- a/lib/protobuf/field/base_field.rb
+++ b/lib/protobuf/field/base_field.rb
@@ -176,7 +176,7 @@ module Protobuf
       def define_accessor(simple_field_name, fully_qualified_field_name)
         message_class.class_eval do
           define_method("#{simple_field_name}!") do
-            @values[fully_qualified_field_name]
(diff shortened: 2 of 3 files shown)


lib/protobuf/message.rb
```diff
@@ -94,7 +94,11 @@ def each_field_for_serialization
     def field?(name)
       field = self.class.get_field(name, true)
       return false if field.nil?
-      @values.key?(field.fully_qualified_name)
+      if field.repeated?
+        @values.key?(field.fully_qualified_name) && !@values[field.fully_qualified_name].empty?
```
nerdrew: nit: You can probably use `.present?` instead of `!blah.empty?`

nerdrew:
LGTM

embark:
@film42 Hoping for a review on this, then will fast-follow with another PR

film42:
Nice bug catch! This is a good fit for 3.7.0 because it subtly changes behavior (though this behavior has been a bug).

I think this is ready to go, but I'm going to ping @liveh2o on this to see what he thinks about this change. I usually ask him for advice about planning breaking changes.

Concise example of the bug:
```ruby
m = ::Test::Resource.new(:name => "thing"); nil
m.repeated_enum! #=> nil -- expected
m.repeated_enum #=> [] -- expected
m.repeated_enum! #=> [] -- this is should be nil
```

Current behavior also considers repeating fields initialized with `[]` to be nil (3.6.X):
```ruby
m = ::Test::Resource.new(:name => "thing", :repeated_enum => []); nil
m.repeated_enum! #=> nil
```

liveh2o:
This is subtle, but ultimately matches the current behavior when initializing an empty repeated enum. I think we're OK moving this forward.

:shipit:

embark:
@film42 thanks for the review! is this good to merge?

film42:
Yes! I'll cut a new pre tomorrow morning. Thanks for the reminders, @embark!

film42:
`3.7.0.pre2` has been released. Thank you!

PR #327, opened 2016-06-09
upcase enums during default value generation

onethirtyfive:
This is an enhancement to #285.

Some of my proto definitions set a starts-with-lowercase enum value as an optional field's default, resulting in problematic code gen.

I'm sorry for the lack of test--if I have a bit more time I'll try to write one. I just want to bring the issue up.

### DIFF
--- a/lib/protobuf/generators/field_generator.rb
+++ b/lib/protobuf/generators/field_generator.rb
@@ -108,7 +108,13 @@ module Protobuf
       private
 
       def enum_default_value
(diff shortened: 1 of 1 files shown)


film42:
Thank you for the contribution! Tests make configuration changes like these easier to digest. If you don't have enough time, we still appreciate your work 😄 .

film42:
Merged here https://github.com/ruby-protobuf/protobuf/pull/366

Thanks @onethirtyfive !!!

PR #329, opened 2016-06-21
Evaluate extension fields in code generation

embark:
This is necessary to access descriptor extensions during code generation. In other PRs, I'll use these "loaded" extension classes later in the code generation to enable custom options.

CC @film42 @liveh2o @nerdrew @zachmargolis 

This one is a little gnarly, but should be the most complicated PR in this series fast followed by the remaining ones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -y libzmq3-dev
   - sudo -E ./install-protobuf.sh
+  # Required for rainbow installation issue, https://github.com/sickill/rainbow/issues/44
+  - gem update --system
   - gem update bundler
 language: ruby
 rvm:
--- a/lib/protobuf/generators/file_generator.rb
+++ b/lib/protobuf/generators/file_generator.rb
@@ -12,7 +12,8 @@ module Protobuf
         super
         @output_file = ::Google::Protobuf::Compiler::CodeGeneratorResponse::File.new(:name => file_name)
         @extension_fields = Hash.new { |h, k| h[k] = [] }
-        @known_messages = []
+        @known_messages = {}
+        @known_enums = {}
         @dangling_messages = {}
       end
 
@@ -49,8 +50,16 @@ module Protobuf
       end
 
       def unknown_extensions
-        @unknown_extensions ||= @extension_fields.reject do |k, _|
-          @known_messages.include?(k)
+        @unknown_extensions ||= @extension_fields.map do |message_name, fields|
+          message_klass = modulize(message_name).safe_constantize
+          if message_klass
+            unknown_fields = fields.reject do |field|
+              @known_messages[message_name] && message_klass.get_field(field.name, true)
+            end
+            [message_name, unknown_fields]
+          else
+            [message_name, fields]
+          end
         end
       end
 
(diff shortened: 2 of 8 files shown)


embark:
CC @film42 @nerdrew @liveh2o the failing test doesn't have anything to do with this PR, so this is ready for review

embark:
Ping @film42

 🐰

lib/protobuf/generators/file_generator.rb
```diff
@@ -142,6 +179,75 @@ def fully_qualified_token?(token)
...
+      end
+
+      def eval_code(context = Object)
+        warn "#{context.inspect}.module_eval #{print_contents.inspect}" if ENV['PB_DEBUG']
+        context.module_eval print_contents.to_s
+        @io.truncate(0)
+        @io.rewind
+      end
+
+      def create_ruby_namespace_heiarchy(name)
```
film42: The code in this method took me a while to understand. I think you can use `rpartition` to achieve the same results:

```
def create_ruby_namespace_heiarchy(namespace)
  loop do
    namespace, _match, _tail = namespace.rpartition(".")
    break if namespace.empty?
    puts namespace
  end
end

irb(main):142:0> create_ruby_namespace_heiarchy('test.pkg.code_generator_spec')
test.pkg
test
=> nil
irb(main):143:0> create_ruby_namespace_heiarchy(".google.protobuf.FieldOptions")
.google.protobuf
.google
```

embark: Seems to work -- I changed the method with you suggestions

embark:
@film42 should we continue to support Ruby 1.9, or allow the json upgrade to drop that support?

embark:
cc @film42 @liveh2o got this back to green. Would be nice to get this merged in -- is there any blocker right now? I could adjust the code structure or add more comments if desired.

.travis.yml
```diff
@@ -2,6 +2,7 @@ before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -y libzmq3-dev
   - sudo -E ./install-protobuf.sh
+  - gem update --system
```
zanker: Is this still necessary?

lib/protobuf/generators/file_generator.rb
```diff
@@ -49,8 +50,14 @@ def compile
       end
 
       def unknown_extensions
-        @unknown_extensions ||= @extension_fields.reject do |k, _|
-          @known_messages.include?(k)
+        @unknown_extensions ||= @extension_fields.map do |message_name, fields|
+          message_klass = modulize(message_name).safe_constantize
+          next [message_name, fields] unless message_klass
```
zanker: Would make this an if statement. What you have right now is a little confusing since using `next` in that fashion isn't that common.

.travis.yml
embark: This is the solution to a rainbow installation issue... dunno when we'll be able to get rid of it

https://github.com/sickill/rainbow/issues/44

lib/protobuf/generators/file_generator.rb
embark: Done

.travis.yml
embark: Added comment to it in case we can remove in the future

film42:
Thanks @embark and @zanker for your help! Sorry this process took so long.

PR #332, opened 2016-07-15
Change the ping port to use ms timeout and make the default 200ms

film42:
Specs are passing locally. Rubocop will fail in this branch.

cc @abrandoned @quixoten


abrandoned:
:shipit:

PR #336, opened 2016-08-17
Remove superfluous rescue from service dispatcher

liveh2o:
Likely a relic of pre-middleware days, the service dispatcher was rescuing all `NoMethodError` exceptions and re-raising it's own `MethodNotFound` exception. The request decoder middleware now handles this case, and it was way too broad (i.e. it was masking actual `NoMethodError` exceptions), so it has been removed.

// @abrandoned @film42

### DIFF
--- a/lib/protobuf/rpc/service_dispatcher.rb
+++ b/lib/protobuf/rpc/service_dispatcher.rb
@@ -28,12 +28,9 @@ module Protobuf
 
       private
 
-      # Call the given service method.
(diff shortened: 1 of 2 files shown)


liveh2o:
I also noticed that some of the Zmq specs (not functional) failed under certain seeds, so we have some order-dependent specs as well. I'll replicate and see if I can sort it out.

abrandoned:
:shipit:

PR #338, opened 2016-09-06
Set maximum version to activesupport gem

vassyz:
Not setting activesupport maximum version will install version 5, which only supports Ruby >= 2.2.2 (released in April 2015). This pull request will currently make bundle install activesupport 4.2.7.1, which also works on  Ruby >= 1.9.3.

### DIFF
--- a/protobuf.gemspec
+++ b/protobuf.gemspec
@@ -19,7 +19,7 @@ require "protobuf/version"
   s.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
   s.require_paths = ["lib"]
 
-  s.add_dependency 'activesupport', '>= 3.2'
+  s.add_dependency 'activesupport', '>= 3.2', '< 5'
   s.add_dependency 'middleware'
   s.add_dependency 'thor'
   s.add_dependency 'thread_safe'


film42:
Will rails 5 work with an activesupport version below 5.0?

vassyz:
Just checked, and it doesn't. I'm not sure what the best way of sorting this out is. I'm guessing protobuf is not the only gem that has this issue.

If somebody is using Rails < 5 and Ruby < 2.2.2 they can lock activesuport to version 4 in their Gemfile to make the protobuf gem work.

    gem 'activesupport', '~> 4.2'

vassyz:
I will close this PR, as I don't think this is the right solution. You need to prioritise Rails 5 users, although bear in mind that the protobuf gem can't be installed at the moment on Ruby < 2.2.2. You could set a required version to 2.2.2, just to let people know.

      s.required_ruby_version     = '>= 2.2.2'  

But the gem works fine on earlier versions, as long as you install activesupport 4.2.

The new version of Bundler (1.13, currently in RC stage) will sort this dependency issue automatically. More info in [this comment](https://github.com/bundler/bundler-features/issues/120#issuecomment-214514847) from André Arko.

PR #344, opened 2016-11-23
Fix "NameError: uninitialized constant Protobuf::Rpc::Protobuf::Socketrpc"

liluo:
Fix "NameError: uninitialized constant Protobuf::Rpc::Protobuf::Socketrpc"

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -56,7 +56,7 @@ Style/TrailingBlankLines:
   Exclude:
     - '**/*.pb.rb'
 
-Style/TrailingComma:
+Style/TrailingCommaInArguments:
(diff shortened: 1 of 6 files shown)


liluo:
@film42

liveh2o:
It looks like the Rubocop upgrade fixed the no method error, but now a bunch of cops are failing. We need the updated Rubocop version, but I don't think we want all of those cops turned on right now. If you want to disable the new cops and update here, we can merge it in. Otherwise, I can update Rubocop in master and you can rebase.

film42:
I agree with @liveh2o . If you wouldn't mind disabling the failing cops, I would love to merge this.

abrandoned:
pulled into master

PR #347, opened 2017-02-11
Abradndoned/extract client server

abrandoned:
in an effort to make our way towards removing client/server code from the core gem and using extension gems to handle these concerns this allows an extension client or server to be loaded instead of requiring the use of Socket or Zmq client/servers

the proposed solution uses `ENV['PB_CLIENT_TYPE']` and `ENV['PB_SERVER_TYPE']` to require and classify a string passed that loads the gem and passes options to the constant the string `classify`'s to

I included specs and left the socket and zmq client and runner in the project (for now) but the intent is to remove them in the next version so they can be loaded from another gem ... to use the embedded client/server with the extension syntax then the rpc server could be launched as follows:

```ruby
PB_SERVER_TYPE="protobuf/rpc/servers/socket_runner" PB_CLIENT_TYPE="protobuf/rpc/connectors/socket" rpc_server --host=127.0.0.1 --port=9999 ./config/environment.rb
```

This structure will allow us to remove the runners and clients from this gem and also experiment with different client/server architectures faster than releasing a new protobuf

@film42 @embark @mmmries

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -56,7 +56,7 @@ Style/TrailingBlankLines:
   Exclude:
     - '**/*.pb.rb'
 
-Style/TrailingCommaInLiteral:
+Style/TrailingComma:
   EnforcedStyleForMultiline: comma
 
 Style/TrivialAccessors:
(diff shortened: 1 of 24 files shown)


quixoten:
LGTM :shipit:

PR #348, opened 2017-02-12
Abrandoned/consolidate base common

abrandoned:
includes everything from #347 but split this out in case of any discussion

consolidates the module `Common` and the class `Base` on the client connectors as they served the same purpose and were for some reason kept separate ... also corrects the interface definition for a client so it will fail if not implemented

@film42


lib/protobuf/rpc/connectors/socket.rb
```diff
@@ -4,7 +4,6 @@ module Protobuf
   module Rpc
     module Connectors
       class Socket < Base
-        include Protobuf::Rpc::Connectors::Common
```
film42: 😍 😍 😍 😍 😍

PR #349, opened 2017-02-16
Fix decoding of packed fields

t-fujii:
Packed repeated field is currently not decoded by #decode of its own class.
The value should be decoded after reading it.


film42:
Looks good to me! @nerdrew @embark would you mind taking a look?

abrandoned:
:shipit:

PR #350, opened 2017-02-17
Add support for custom file options

embark:
cc @film42 @liveh2o @zanker @nerdrew

File Options!

The longest of the option PRs, as there is some boiler plate to set up.


abrandoned:
:shipit:

PR #352, opened 2017-02-20
Bugfix: enum_for_tags returns nil if tag nil

evanboho:
This fixes https://github.com/ruby-protobuf/protobuf/issues/265
Bug: calling to_i on nil returns 0, and so will return an erroneous
value if the enum is zero-indexed.


liveh2o:
Looks good to me. :shipit:

PR #356, opened 2017-03-01
pull in acceptable? optimization for varint and fix rubocop for previ…

abrandoned:
…ously merged PR

pulls in varint "acceptable?" check from #297 

also fixes the previously merged rubocop issue in base_spec.rb

@film42


film42:
Looks good. Just waiting on travis.

PR #361, opened 2017-03-04
acceptable check is not needed most places as coerce runs the same lo…

abrandoned:
…gic and use fetch instead of checking for key

most field types do acceptability check in `coerce!` and in `acceptable?`, this leaves acceptable but doesn't use it in the `set_field` method as the `coerce!` call will take care of that for us

also uses Hash `fetch` instead of doing a `key?` check before access as the default can be passed into the fetch and again avoid another conditional check

@film42

### DIFF
--- a/lib/protobuf/field/varint_field.rb
+++ b/lib/protobuf/field/varint_field.rb
@@ -54,8 +54,10 @@ module Protobuf
         int_val = if val.is_a?(Integer)
                     return true if val >= 0 && val < INT32_MAX # return quickly for smallest integer size, hot code path
                     val
+                  elsif val.is_a?(Numeric)
+                    val.to_i
                   else
-                    coerce!(val)
+                    Integer(val, 10)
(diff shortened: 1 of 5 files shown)


lib/protobuf/field/varint_field.rb
```diff
@@ -64,8 +66,11 @@ def acceptable?(val)
       end
 
       def coerce!(val)
+        fail TypeError, "Expected value of type '#{type_class}' for field #{name}, but got '#{val.class}'" unless acceptable?(val)
         return val.to_i if val.is_a?(Numeric)
         Integer(val, 10)
+      rescue
+        fail TypeError, "Expected value of type '#{type_class}' for field #{name}, but got '#{val.class}'"
```
film42: Can we just catch the specific error and reraise? I don't love catch all rescues because it's hard to detect other things breaking.

PR #366, opened 2017-03-09
Upcase enum default

embark:
This is just an extension of https://github.com/ruby-protobuf/protobuf/pull/327 with a test

Thanks @onethirtyfive for the contribution!

cc @film42 @nerdrew


nerdrew:
LG

PR #368, opened 2017-04-27
Use alias_method in place of alias_method_chain

embark:
Today activesupport 5.1.0 was released which no longer supports alias_method_chain. It was deprecated prior to this version. cc @nerdrew @film42 @abrandoned

### DIFF
--- a/lib/protobuf/code_generator.rb
+++ b/lib/protobuf/code_generator.rb
@@ -86,7 +86,8 @@ module Protobuf
         end
       end
 
-      alias_method_chain :set, :options
+      alias_method :set_without_options, :set
+      alias_method :set, :set_with_options
     end
   end
 end


embark:
Oops prepend not supported before ruby 2.0, will switch to regular alias_method

embark:
These rpc related test failures look like possibly another gem update broke them. But all the related tests are passing... :)

film42:
Looks good!

Just going to look at what broke to figure out if we need to make changes or lock some things down.

film42:
I pulled down zmq 3.2.2 and can confirm it's the problem. We should be using 3.2.5. Looks like travis has their own cached version.

I'll merge this and open a PR to fix travis. Once that's passing and merged, I'll go ahead and release.

Thanks again, @embark !

film42:
Actually needed v4 but it's working now. Released as `3.7.1`.

embark:
@film42 thanks for the fix :) I've never gotten the zmq stuff running on my machine.

PR #369, opened 2017-04-28
Travis caches an old zmq version. This uses the latest version from 4

film42:
stable.

Weird, right? You can see we should be using v4 here: https://packages.debian.org/search?keywords=libzmq3-dev

cc @abrandoned @mmmries @embark @quixoten


mmmries:
:shipit:

PR #371, opened 2017-05-03
Override stats#server from the response

quixoten:
This ensures that the server fulfilling a request will correctly end up in the logs.

@abrandoned @film42

### DIFF
--- a/lib/protobuf/rpc/connectors/base.rb
+++ b/lib/protobuf/rpc/connectors/base.rb
@@ -104,6 +104,7 @@ module Protobuf
           # Parse out the raw response
           @stats.response_size = @response_data.size unless @response_data.nil?
           response_wrapper = ::Protobuf::Socketrpc::Response.decode(@response_data)
+          @stats.server = response_wrapper.server if response_wrapper.field?(:server)
 
           # Determine success or failure based on parsed data
           if response_wrapper.field?(:error_reason)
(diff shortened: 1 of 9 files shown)


lib/protobuf/rpc/connectors/base.rb
```diff
@@ -104,6 +104,7 @@ def parse_response
           # Parse out the raw response
           @stats.response_size = @response_data.size unless @response_data.nil?
           response_wrapper = ::Protobuf::Socketrpc::Response.decode(@response_data)
+          @stats.server = response_wrapper.server if response_wrapper.field?(:server)
```
film42: This will always be true since they need to have the latest version with the new response proto to exec this code, right?

quixoten: A user could override there config to use `nil` as the server.

quixoten: :thinking: but if they do set it to `nil` the current implementation will set it back to `Socket.gethostname`. Should probably fix that.

film42: What about `@stats.server = response_wrapper.server || "SERVER_HOSTNAME_NOT_PROVIDED"`?

quixoten:
This is ready for another review. I reverted changes to the server side code. I want to put up a couple PRs for those changes using different implementations, because I wasn't crazy about my initial direction. This should get everything back in a releasable state. The effective changes since the last release can be seen here: https://github.com/ruby-protobuf/protobuf/compare/27bb8e29b4adf014b62cb7276abf41cf0ba9b3db...logserver

film42:
:shipit:

PR #375, opened 2017-05-17
reload context causes depletion of file descriptors

abrandoned:
causing issues with depletion of file descriptors

@film42 @quixoten


quixoten:
:shipit:

PR #367, opened 2017-03-29
Map types

embark:
Support for map types. Primary commit 9dc224e was written by @jhump, I'm merely a proxy :)

cc @nerdrew @film42 @abrandoned @zanker

PR #377, opened 2017-07-23
Add option to have require_relative statements

tdeo:
To account for this issue: https://github.com/ruby-protobuf/protobuf/issues/240. 

I chose to use an environment variable `REQUIRE_RELATIVE=true` as I can't see another way to pass options to the gem

### DIFF
--- a/lib/protobuf/generators/file_generator.rb
+++ b/lib/protobuf/generators/file_generator.rb
@@ -125,7 +125,7 @@ module Protobuf
         header "Imports"
 
         descriptor.dependency.each do |dependency|
-          print_require(convert_filename(dependency))
+          print_require(convert_filename(dependency), ENV.key?('PB_REQUIRE_RELATIVE'))
         end
 
         puts
--- a/lib/protobuf/generators/printable.rb
+++ b/lib/protobuf/generators/printable.rb
@@ -127,8 +127,8 @@ module Protobuf
       #
       #   print_require('foo/bar/baz') -> "require 'foo/bar/baz'"
       #
-      def print_require(file)
-        puts "require '#{file}'"
+      def print_require(file, relative = false)
+        puts "require#{'_relative' if relative} '#{file}'"
       end
 
       # Puts the given message prefixed by the indent level.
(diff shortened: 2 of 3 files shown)


tdeo:
@film42 let me know if that's a good solution and I'll add instructions in the readme for that.

lib/protobuf/generators/file_generator.rb
```diff
@@ -125,7 +125,7 @@ def print_import_requires
         header "Imports"
 
         descriptor.dependency.each do |dependency|
-          print_require(convert_filename(dependency))
+          print_require(convert_filename(dependency), ENV['REQUIRE_RELATIVE'] == 'true')
```
film42: We've been prefacing env vars with `PB_`. Would you mind doing that here? `PB_REQUIRE_RELATIVE`?

film42:
Sorry for the delay in getting this reviewed btw! Thanks for your contribution ❤️ !

lib/protobuf/generators/file_generator.rb
tdeo: Sure

tdeo:
I just changed the code to change the environment variable name to `PB_REQUIRE_RELATIVE` and to only be checked for presence as the others. I'll edit the wiki when it gets merged.

Thanks

tdeo:
@film42 Do you have any idea when you'll be able to merge this / release ? Thanks

film42:
@tdeo I'll release this right after we get map types released #367.

ojab:
I've almost sent a PR that adds option to use `require_dependency` instead (it helps with Rails autoload + protobuf imports), but I assume that it's better to tweak this PR and have just one option (something like `PB_REQUIRE_KEYWORD` that can be `require_relative`, `require_dependency` or whatever else).
Sounds good, can be done?

And, btw, `print_require` is used in [print_generic_requires](https://github.com/ruby-protobuf/protobuf/blob/bbdeae031dccae8ad1f4c8c2ed39b7aad6b1f9aa/lib/protobuf/generators/file_generator.rb#L116), I think that it'll break with `require_relative` and alike.

ojab:
Ah, please ignore statement about `print_generic_requires`, after rereading the patch I see that `relative` is passed only for imports.

film42:
Released in `v3.8.0`! Thanks for the contribution!

PR #378, opened 2017-07-24
use a modern fork of the middleware gem

nerdrew:
cc @embark

### DIFF
--- a/protobuf.gemspec
+++ b/protobuf.gemspec
@@ -22,7 +22,7 @@ require "protobuf/version"
   # Hack, as Rails 5 requires Ruby version >= 2.2.2.
   active_support_max_version = "< 5" if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.2.2")
   s.add_dependency "activesupport", '>= 3.2', active_support_max_version
-  s.add_dependency 'middleware'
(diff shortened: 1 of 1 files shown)


nerdrew:
The middleware gem looks unmaintained. This gem is a fork that has merged helpful PRs. https://github.com/Ibsciss/ruby-middleware

embark:
Nice, thanks!

film42:
I'm pretty hesitant to merge something like this since our team has a number of projects using the existing middleware gem. I wish the maintainer had renamed it to something else. I imagine this would introduce a number of conflicts depending on which gem loads first.

nerdrew:
Yes, the gems are mutually exclusive. The new gem is a superset of functionality it seems. The old gem seems to disregard return values which seems unintuitive for a middleware stack.

film42:
Yeah, that's unintuitive, but we can't swap out a dependency with a conflicting name. That's a breaking change with zero discoverability. A while back I forked `middleware` to support variadic args and switched the name to avoid a collision: https://github.com/film42/midler. If this were the case, we could weight pros and cons, but since it's not the case, I can't merge this.

Thanks for opening a PR though @nerdrew . I'm glad someone is trying to maintain middleware.

PR #380, opened 2017-10-06
Fix compatibility issues for Ruby 2.4.0

RickCSong:
This change uses 1.class instead of hard coding to Fixnum to prevent
Ruby >= 2.4.0 from raising a deprecation warning. We use this
implementation to also guarantee backward compatibility as well.


film42:
Thanks! Would you mind adding `2.4` to the list in `.travis.yml` so we can begin looking at ruby 2.4 moving forward?

RickCSong:
@film42 Sorry for the delay!  This PR totally fell off my radar.  Updated and added 2.4 to travis.

Issue #341, opened 2016-10-21
--ruby-out does not invoke protoc-gen-ruby for protoc v3


ebenoist:
Hi Friends-

With the protobuf v3 out, the official google compiler invokes its internal ruby compiler for v3 protof files with the `--ruby_out` option instead of protoc-gen-ruby. I'm not ready to switch to v3 yet (or Google's compiler) and would love to continue using this project.

I can force this to work by renaming the flag to `--ruby2_out` for example and by changing the name of the binary to `protoc-gen-ruby2` this works fantastically even with the newest compiler. Any thoughts on how both of these compilers can continue to live together. This project is invaluable because of its proto2 support.

If there's consensus around an approach, I'd be happy to put up a PR.

brianvh:
I haven attempted to replicate these steps, using the current 3.6.12 release, and all I get is an immediate Exception, from the Gem::Specification layer:

```
[hd]: protoc --ruby2_out=lib -I definitions definitions//foo.proto
/Users/brianvh/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/rubygems/specification.rb:2158:in `method_missing': undefined method `this' for #<Gem::Specification:0x3fc3d049b424 protobuf-3.6.12> (NoMethodError)
	from /Users/brianvh/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/rubygems/specification.rb:1057:in `find_active_stub_by_path'
	from /Users/brianvh/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:64:in `require'
	from /Users/brianvh/.rvm/gems/ruby-2.3.0@hr-api/gems/protobuf-3.6.12/bin/protoc-gen-ruby:9:in `<top (required)>'
	from /Users/brianvh/.rvm/gems/ruby-2.3.0@hr-api/bin/protoc-gen-ruby2:23:in `load'
	from /Users/brianvh/.rvm/gems/ruby-2.3.0@hr-api/bin/protoc-gen-ruby2:23:in `<main>'
	from /Users/brianvh/.rvm/gems/ruby-2.3.0@hr-api/bin/ruby_executable_hooks:15:in `eval'
	from /Users/brianvh/.rvm/gems/ruby-2.3.0@hr-api/bin/ruby_executable_hooks:15:in `<main>'
--ruby2_out: protoc-gen-ruby2: Plugin failed with status code 1.
```

[...]

PR #382, opened 2018-02-22
Update the compile task to work with protoc v3

liveh2o:
The Protocol Buffers v3 compiler (protoc) now comes with a builtin Ruby plugin. This conflicts with our local custom Ruby plugin. In order to use it, the custom plugin must be passed to the compiler, and invoked via the custom output generator:

```
$ protoc --plugin=protoc-gen-ruby-protobuf=/path/to/bin/protoc-gen-ruby --ruby-protobuf_out=lib
```

Update the compile task to take this new pattern into account so that it works with both the v2 and v3 compilers.

// @film42 @brianstien

### DIFF
--- a/lib/protobuf/tasks/compile.rake
+++ b/lib/protobuf/tasks/compile.rake
@@ -1,13 +1,32 @@
-require 'fileutils'
+require "fileutils"
 
 namespace :protobuf do
 
-  desc 'Clean & Compile the protobuf source to ruby classes. Pass PB_NO_CLEAN=1 if you do not want to force-clean first.'
+  desc "Clean & Compile the protobuf source to ruby classes. Pass PB_NO_CLEAN=1 if you do not want to force-clean first."
   task :compile, [:package, :source, :destination, :plugin, :file_extension] do |_tasks, args|
-    args.with_defaults(:destination => 'lib')
-    args.with_defaults(:source => 'definitions')
-    args.with_defaults(:plugin => 'ruby')
-    args.with_defaults(:file_extension => '.pb.rb')
+    binpath = ::File.expand_path("../../../../bin", __FILE__)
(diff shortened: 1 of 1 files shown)


lib/protobuf/tasks/compile.rake
```diff
@@ -1,13 +1,32 @@
...
 namespace :protobuf do
 
-  desc 'Clean & Compile the protobuf source to ruby classes. Pass PB_NO_CLEAN=1 if you do not want to force-clean first.'
+  desc "Clean & Compile the protobuf source to ruby classes. Pass PB_NO_CLEAN=1 if you do not want to force-clean first."
   task :compile, [:package, :source, :destination, :plugin, :file_extension] do |_tasks, args|
-    args.with_defaults(:destination => 'lib')
-    args.with_defaults(:source => 'definitions')
-    args.with_defaults(:plugin => 'ruby')
-    args.with_defaults(:file_extension => '.pb.rb')
+    binpath = ::File.expand_path("../../../../bin", __FILE__)
```
liveh2o: If there is a better way to do this, I'm open to suggestions.

liveh2o:
Resolves #341.

lib/protobuf/tasks/compile.rake
film42: I think it's much better than searching local path, so :+1:

mvz:
The wiki mentions the old method of compiling that no longer works. Unfortunately, I don't know enough yet to update it.

lib/protobuf/tasks/compile.rake
mvz: @liveh2o How about `Gem.bin_path('protobuf', 'protoc-gen-ruby')`?

PR #383, opened 2018-05-03
Base64 encode byte fields when converting to JSON

drcapulet:


film42:
Looks great! Thank you, @drcapulet !

PR #385, opened 2018-06-07
Implement Enum#to_json to

mattnichols:
There is an issue in JrJackson RubyAnySerializer that does not render Protobuf::Enum correctly because it does not implement to_json. This fixes that issue here until the serializer can be patched.

@abrandoned @ah @brianstien

### DIFF
--- a/lib/protobuf/enum.rb
+++ b/lib/protobuf/enum.rb
@@ -288,7 +288,7 @@ module Protobuf
     #     render Protobuf enums correctly because to_json is not defined.
     #
     def to_json
-      to_s
+      to_i
     end
 
     def to_s(format = :tag)
--- a/lib/protobuf/enum.rb
+++ b/lib/protobuf/enum.rb
@@ -283,6 +283,14 @@ module Protobuf
       tag.to_int
     end
 
+    ##
+    # This fixes a reflection bug in JrJackson RubyAnySerializer that does not
+    #     render Protobuf enums correctly because to_json is not defined.
+    #
+    def to_json
+      to_s
+    end
+
     def to_s(format = :tag)
       case format
       when :tag
--- a/spec/lib/protobuf/enum_spec.rb
+++ b/spec/lib/protobuf/enum_spec.rb
@@ -162,7 +162,7 @@ RSpec.describe Protobuf::Enum do
 
     describe '.to_json' do
       it 'renders the enum value' do
-        expect(Test::EnumTestType::ONE.to_json).to eq("1")
+        expect(Test::EnumTestType::ONE.to_json).to eq(1)
         expect({ :value => Test::EnumTestType::ONE }.to_json).to eq(%({"value":1}))
       end
     end
--- a/spec/lib/protobuf/enum_spec.rb
+++ b/spec/lib/protobuf/enum_spec.rb
@@ -160,6 +160,13 @@ RSpec.describe Protobuf::Enum do
       end
     end
 
+    describe '.to_json' do
+      it 'renders the enum value' do
+        expect(Test::EnumTestType::ONE.to_json).to eq("1")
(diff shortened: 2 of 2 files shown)


mattnichols:
@brianstien 's simple repro of the problem in JrJackson:

```
jruby-9.1.12.0 :001 > class Foo < SimpleDelegator; end
 => nil 
jruby-9.1.12.0 :002 > JrJackson::Json.dump(:foo => Foo.new(1))
 => "{\"foo\":\"#<Foo>\"}"
```

```
# defining to_json will fix this, or should we define to_json_data?
jruby-9.1.12.0 :003 > class Foo < SimpleDelegator; def to_json; __getobj__.to_json; end; end
 => :to_json 
jruby-9.1.12.0 :004 > JrJackson::Json.dump(:foo => Foo.new(1))
 => "{\"foo\":1}"
```

abrandoned:
@mattnichols I think this same functionality is accounted for in the other PR open https://github.com/ruby-protobuf/protobuf/pull/383

we should add the spec and see if we can merge in this spec with the other branch and make it more comprehensive

film42:
@mattnichols I merged #383 and rebased against master for you. Things are passing locally, so if travis is good to go, then I think we're ready to :rocket: .

brianstien:
@film42 Can you remove your commit please? We did need that to_json method defined.

@mattnichols I think to_json should be `to_i`, not `to_s`

mattnichols:
@film42 This is the code that is requires the explicit `.to_json` implementation. https://github.com/guyboertje/jrjackson/blob/master/src/main/java/com/jrjackson/RubyAnySerializer.java#L37

This is the behavior with your code changes (minus the `def to_json`)

```ruby
synchronicity#development >> require "protobuf"
=> false
synchronicity#development >> require "jrjackson"
=> false
synchronicity#development >>
synchronicity?                 class EnumTest < ::Protobuf::Enum; define :ONE, 1; end
=> #<Protobuf::Enum(EnumTest)::ONE=1>
synchronicity#development >> { :value => EnumTest::ONE }.to_json
=> "{\"value\":1}"
synchronicity#development >>
synchronicity?                 class Foo < SimpleDelegator; end
=> nil
synchronicity#development >> JrJackson::Json.dump(:foo => EnumTest::ONE)
=> "{\"foo\":\"#<EnumTest>\"}"
```

film42:
@mattnichols Excellent! I was thinking it was a `to_json` bug on the root object, not the individual fields, but that makes sense.

@brianstien I removed my commit, and since I was already pushing, made the `to_i` change as well.

brianstien:
@film42 That should do it, thanks!

mattnichols:
```ruby
synchronicity#development >> require "protobuf"
=> false
synchronicity#development >> require "jrjackson"
=> false
synchronicity#development >> class EnumTest < ::Protobuf::Enum; define :ONE, 1; end
=> #<Protobuf::Enum(EnumTest)::ONE=1>
synchronicity#development >> { :value => EnumTest::ONE }.to_json
=> "{\"value\":1}"
synchronicity#development >> class Foo < SimpleDelegator; end
=> nil
synchronicity#development >> JrJackson::Json.dump(:foo => EnumTest::ONE)
=> "{\"foo\":1}"
```

Appears to be working now

film42:
Looks like travis is struggling. I'm running `2.5.1` locally and specs are passing:
```
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Protobuf::Field.build 
     # Not yet implemented
     # ./spec/lib/protobuf/field_spec.rb:8

Finished in 20.07 seconds (files took 0.82061 seconds to load)
659 examples, 0 failures, 1 pending
```

film42:
I think this is good to go. @brianstien ? @abrandoned ?

abrandoned:
:shipit:

PR #386, opened 2018-06-11
Fix enum's to_json

drcapulet:
This PR fixes two regressions from #385.

First, `to_json` needs to return a string:
```ruby
irb(main):003:0> 5.to_json
=> "5"
```

Second, when using `JSON.dump`, `to_json` is expected to take a `JSON::State` argument - see ActiveSupport's comment [here](https://github.com/rails/rails/blob/b5aa0266d0a4d5f7a02eb31df730fa172d1a845a/activesupport/lib/active_support/core_ext/object/json.rb#L34-L46).

### DIFF
--- a/lib/protobuf/enum.rb
+++ b/lib/protobuf/enum.rb
@@ -283,12 +283,13 @@ module Protobuf
       tag.to_int
     end
 
-    ##
     # This fixes a reflection bug in JrJackson RubyAnySerializer that does not
-    #     render Protobuf enums correctly because to_json is not defined.
+    # render Protobuf enums correctly because to_json is not defined. It takes
+    # any number of arguments to support the JSON gem trying to pass an argument.
+    # NB: This method is required to return a string and not an integer.
     #
-    def to_json
-      to_i
+    def to_json(*)
+      to_s
     end
 
     def to_s(format = :tag)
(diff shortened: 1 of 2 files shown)


lib/protobuf/enum.rb
```diff
@@ -283,12 +283,13 @@ def to_int
...
     # This fixes a reflection bug in JrJackson RubyAnySerializer that does not
-    #     render Protobuf enums correctly because to_json is not defined.
+    # render Protobuf enums correctly because to_json is not defined. It takes
+    # any number of arguments to support the JSON gem trying to pass an argument.
+    # NB: This method is required to return a string and not an integer.
     #
-    def to_json
-      to_i
+    def to_json(*)
+      to_s
```
film42: We actually debated this and realized that it would make more sense for this to be an integer. Perhaps I should have released this as a minor version bump. Were you previously relying on a string value from this method?

drcapulet: It’s the method signature required by the JSON gem - if you checkout the branch, switch it back to to_i, and run the specs you’ll get a TypeError.

film42:
Thanks @drcapulet ! I'll look at the `to_i` stuff later.

PR #390, opened 2018-11-09
Fix enum hash equality

liveh2o:
Protobuf::Enum delegates methods to Fixnum, which has a custom hash equality method (`eql?`). This causes enum values to be equivalent when using `==`, `===`, `equals?`, but not `eql?`**:

```
2.3.7 :002 > Test::EnumTestType::ZERO.eql?(::Test::EnumTestType::ZERO)
 => false
```

However, they are equilvalent to their tag value:

```
2.3.7 :002 > Test::EnumTestType::ZERO.eql?(::Test::EnumTestType::ZERO.tag)
 => true
```

Use the hash equality implementation from `Object#eql?`, which is equivalent to `==` instead.

**The implementation changed in Ruby 2.5, so this only affects Ruby versions less than v2.5.

### DIFF
--- a/lib/protobuf/enum.rb
+++ b/lib/protobuf/enum.rb
@@ -271,6 +271,25 @@ module Protobuf
       tag.class
     end
 
+    # Protobuf::Enum delegates methods to Fixnum, which has a custom hash equality method (`eql?`)
+    # This causes enum values to be equivalent when using `==`, `===`, `equals?`, but not `eql?`**:
+    #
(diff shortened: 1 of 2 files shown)


film42:
Released as `3.8.5.pre0` for testing purposes.

liveh2o:
Released in [v3.8.5](https://github.com/ruby-protobuf/protobuf/releases/tag/v3.8.5).

PR #391, opened 2018-12-31
Bdewitt/remove helpers

abrandoned:
Quite a few optimizations for MRI and JRuby; started with too large a cut (including some protobuf java helpers for varint encoding) but paired back to just including optimizations in the first PR. Will be starting another gem to include the varint changes for java (which yields another significant increase in perf for encode/decode)

@film42 @quixoten @brianstien @liveh2o 

Primary optimizations were removing the need to do runtime checking of `map?`/`repeated?` on field objects by writing the access methods `set/set_field/value_from_values` directly onto the field object; also sped up the access of each `get_field` call by using a constant dynamically defined on inherited object, JRuby has much faster access to Constants than class method lookups; also removed superfluous lookups of fields and `respond_to` calls to be replaced by methods on the field object

included the benchmark I ran as `varint_prof.rb` but will probably move it into a rake task so we can use it for future optimization; will also squash all the commits for a final merge but keeping them in the branch for now as changes are discussed

Benchmark/ips if encode/decode/to_hash on a protobuf message with MRI and JRuby below

JRuby -- approx 5x faster `to_hash`
               2.5x faster `encode`
               1.5x faster `decode`
```bash
# BEFORE PR
JRUBY_OPTS="--server --disable:did_you_mean" bundle exec ruby ./varint_prof.rb
Warming up --------------------------------------
             to_hash    32.455k i/100ms
              encode    15.644k i/100ms
              decode    18.291k i/100ms
Calculating -------------------------------------
             to_hash    394.430k (± 1.5%) i/s -      7.886M in  19.999530s
              encode    175.397k (± 1.8%) i/s -      3.520M in  20.075313s
              decode    209.607k (± 2.2%) i/s -      4.207M in  20.081145s

# AFTER PR applied
JRUBY_OPTS="--server --disable:did_you_mean" bundle exec ruby ./varint_prof.rb 
Warming up --------------------------------------
             to_hash   102.823k i/100ms
              encode    33.450k i/100ms
              decode    26.993k i/100ms
Calculating -------------------------------------
             to_hash      1.934M (± 4.1%) i/s -     38.661M in  20.034220s
              encode    443.646k (± 3.6%) i/s -      8.864M in  20.011563s
              decode    351.769k (± 2.2%) i/s -      7.045M in  20.038938s
```

MRI -- approx 3x faster `to_hash`
           1.3x faster `encode`
           1.2x faster `decode`
```bash
# BEFORE
bundle exec ruby ./varint_prof.rb
Warming up --------------------------------------
             to_hash    36.539k i/100ms
              encode    11.728k i/100ms
              decode    10.526k i/100ms
Calculating -------------------------------------
             to_hash    419.455k (± 0.7%) i/s -      8.404M in  20.036460s
              encode    120.492k (± 2.8%) i/s -      2.416M in  20.068034s
              decode    109.387k (± 0.7%) i/s -      2.189M in  20.016147s

# AFTER PR applied
bundle exec ruby ./varint_prof.rb 
Warming up --------------------------------------
             to_hash   105.286k i/100ms
              encode    15.569k i/100ms
              decode    12.675k i/100ms
Calculating -------------------------------------
             to_hash      1.402M (± 0.8%) i/s -     28.111M in  20.058605s
              encode    167.251k (± 1.2%) i/s -      3.347M in  20.016641s
              decode    134.016k (± 0.8%) i/s -      2.687M in  20.052077s
```


liveh2o:
Overall, this looks good. I especially appreciate getting rid of that crazy `set_field` method with all of the nested conditionals. 💯

lib/protobuf/field/base_field_method_definitions.rb
```diff
@@ -0,0 +1,371 @@
+module Protobuf
+  module Field
+    class BaseFieldMethodDefinitions
+
+      def self.fully_qualified_name_string(selph)
+        fully_qualified_name = ""
+        fully_qualified_name << ":"
+        fully_qualified_name << '"' if selph.fully_qualified_name.to_s.start_with?(".")
```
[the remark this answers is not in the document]
abrandoned: I don't think adding an abstraction inside a method that is called to define a method is necessary

lib/protobuf/field/base_field_method_definitions.rb
```diff
@@ -0,0 +1,371 @@
...
+        fully_qualified_name << selph.fully_qualified_name.to_s
+        fully_qualified_name << '"' if selph.fully_qualified_name.to_s.start_with?(".")
+        fully_qualified_name
+      end
+
+      def self.define_to_hash_value_to_message_hash!(selph)
```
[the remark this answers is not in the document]
liveh2o: I meant to say "**not** an instance of...". 😞 Updated the comment.

So you are saying that it is a reference to the field class that the methods are being added to?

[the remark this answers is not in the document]
abrandoned: yes, it's `self` in terms of the object he methods are being defined on; but they are singleton methods, we can't apply the methods to the Class because they behave differently based on the instance of the object

lib/protobuf/field/int64_field.rb
```diff
@@ -16,6 +16,19 @@ def self.min
...
+      ##
+      # Instance Methods
+      #
+      def acceptable?(val)
+        if val.is_a?(Integer) || val.is_a?(Numeric)
+          val >= INT64_MIN && val <= INT64_MAX
+        else
+          Integer(val, 10) >= INT64_MIN && Integer(val, 10) <= INT64_MAX
+        end
+      rescue
```
film42: I think we should explicitly catch the `ArgumentError` (vs rescue all) to avoid accidental typos slipping in.

lib/protobuf/message.rb
```diff
@@ -80,37 +70,26 @@ def each_field
...
-          array = Array.new(value.size)
-          i = 0
-          value.each do |k, v|
-            array[i] = field.type_class.new(:key => k, :value => v)
-            i += 1
-          end
-          value = array
-        end
+      _protobuf_message_required_field_tags.each do |tag|
+        fail ::Protobuf::SerializationError, "Required field #{self.class.name}##{_protobuf_message_field[tag].name} does not have a value."
```
film42: Should this have a conditional check before raising an error?

lib/protobuf/field/int64_field.rb
abrandoned: well, my intent was to not change anything except the serialization methods .... we can make this change but it would change how today's code works

lib/protobuf/message.rb
abrandoned: what conditional check would you recommend? only the required field tags are in the array so if anything is present in the array then it's a problem

film42: I see now. `_protobuf_message_required_field_tags` only stores the unset field tags. I think it would be easier to understand if the function mentioned it only contains unset attributes.

lib/protobuf/field/int64_field.rb
film42: OK. We can punt this change for now.

lib/protobuf/field/base_field.rb
```diff
@@ -93,6 +103,68 @@ def default_value
...
+        elsif message? || type_class == ::Protobuf::Field::BytesField
+          ::Protobuf::Field::BaseFieldMethodDefinitions.define_bytes_encode_to_stream_method!(self)
+        elsif type_class == ::Protobuf::Field::StringField
+          ::Protobuf::Field::BaseFieldMethodDefinitions.define_string_encode_to_stream_method!(self)
+        else
+          ::Protobuf::Field::BaseFieldMethodDefinitions.define_base_encode_to_stream_method!(self)
+        end
+      end
+
+      def define_field_p!
```
film42: I think these `_p` methods would read a lot better with `_predicate` spelled out. I'll leave it up to you.

film42:
I've read through this PR several times now and it looks good to me. If specs are passing I think we're good to cut a pre-release. Nice work, @abrandoned !

PR #392, opened 2019-02-04
WIP: Adding tracing

minond:
Adding service to service tracing using Opentracing. This is still a WIP, item left to do are the following:

- [ ] Switch to proper encoding method
- [ ] Gather and include metrics


abrandoned:
@minond I think this is better as a gem that injects a middleware, any reason we don't take that approach?

minond:
Closing this PR and getting another one up using the middleware feedback.

PR #398, opened 2019-06-25
Alias #eql? to #== and define #hash on Protobuf::Message

drcapulet:
Hashes (and some methods of Array) use the `#eql?` and `#hash` methods, this defines them so that those operations work correctly.


drcapulet:
Ping @film42, anything else needed here?

PR #404, opened 2019-10-10
Server before/afer startup callbacks

film42:
We currently have the ability to subscribe to callbacks via:

```
ActiveSupport::Notifications.subscribe('after_server_bind') { ... }
``` 

This proposes 2 changes:

1. Add a new `before_server_bind` event that is published before a runner's `run` method is called via the CLI.
2. A new helper method for {before,after}_server_bind on the Protobuf module.

Example:

```ruby
Protobuf.before_server_bind do
  eager_load_some_data_only_when_server_starts!
end

Protobuf.after_server_bind do
  report_server_online!
end
```

I also considered adding these to the `Protobuf::Rpc::Server` module since it's more specific to the use-case, but felt like the root `Protobuf`  was "good enough" so please let me know if you have a strong preference either way.

cc @liveh2o @abrandoned

### DIFF
--- a/lib/protobuf.rb
+++ b/lib/protobuf.rb
@@ -54,6 +54,18 @@ module Protobuf
     attr_writer :client_host
   end
 
+  def self.after_server_bind(&block)
+    ::ActiveSupport::Notifications.subscribe('after_server_bind') do |*args|
+      block.call(*args)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


liveh2o:
Not sure what’s up with the build failing.

PR #406, opened 2019-11-30
Fix code generation in spec task to work with proto3

mvz:
Taken from 8e7c96a820ac42783611a1e72ada6adc93fade7d


mvz:
A similar change was already applied in master so this pull request is no longer relevant.

PR #407, opened 2019-12-01
Update TravisCI build

mvz:
Builds on newer protobuf versions while not allowing failures.

Follow-up for #406, which made the build for protobuf 3.0.0 green.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ AllCops:
   Exclude:
     - 'spec/support/protos/*.pb.rb'
     - 'varint_prof.rb'
+    - 'protobuf-*/**/*.rb'
 
 Lint/EndAlignment:
   AlignWith: keyword
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ rvm:
   - rbx-2
 env:
   - PROTOBUF_VERSION=2.6.1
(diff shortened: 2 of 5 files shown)


olleolleolle:
@mvz Does it make sense to rebase this? Does it still work?

mvz:
I don't think much has changed but I'll give it a shot.

mvz:
It looks like there's a problem with pry and pry-byebug versions.

deivid-rodriguez:
Yeah.

On ruby 2.3, `pry-byebug` gets resolved to a version older than the latest (3.7.0) because the latest does not support ruby 2.3. But `pry` gets resolved to `0.13` which is not supported by `pry-byebug` 3.7.0, only by the latest version.

Possible solutions:
* Pin `pry` to an older version.
* Drop support for ruby 2.3.

mvz:
It looks like the maintainers of this gem want to still target 2.3 (see #412), so I'll go with the first option.

olleolleolle:
@mvz The `rbx-2` needs to run in `dist: trusty` to even begin. If you're thinking of keeping it.

```yaml
matrix:
  include:
    - rvm: rbx-2
      dist: trusty
```

mvz:
I think I'll wait for #406 and perhaps #412 to be merged and then see again which parts of this pull request are still relevant.

mvz:
Closing this since this project no longer uses TravisCI.

PR #411, opened 2020-01-06
Add Message.from_json

aslakhellesoy:
This PR adds a new `Message.from_json` class method that turns a JSON string into a message object.

It addresses two shortcomings:

* Lower camel case field names (`fooBarZap`) in the JSON are converted to underscore (`fooBarZap`)
* Values for `bytes` fields are Base64 decoded unless they are already in "binary" format. This is because Protobuf's JSON representation uses Base64 encoding for bytes fields.

### DIFF
--- a/lib/protobuf/field/bytes_field.rb
+++ b/lib/protobuf/field/bytes_field.rb
@@ -48,7 +48,15 @@ module Protobuf
 
       def coerce!(value)
         case value
-        when String, Symbol
+        when String
+          if value.encoding == Encoding::ASCII_8BIT
+            # This is a "binary" string
+            value
+          else
+            # Assume the value is Base64 encoded (from JSON)
+            Base64.decode64(value)
+          end
+        when Symbol
           value.to_s
         when NilClass
           nil
--- a/lib/protobuf/message.rb
+++ b/lib/protobuf/message.rb
@@ -21,6 +21,22 @@ module Protobuf
       name
     end
 
+    def self.from_json(json)
+      fields = normalize_json(JSON.parse(json))
+      self.new(fields)
+    end
+
+    def self.normalize_json(ob)
+      case ob
+      when Array
+        ob.map {|value| normalize_json(value) }
+      when Hash
+        Hash[*ob.flat_map {|key, value| [key.underscore, normalize_json(value)] }]
+      else
+        ob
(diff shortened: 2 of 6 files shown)


aslakhellesoy:
Please hold off on merging this while I fix the failing tests. I'm also adding a bugfixes for enum fields and JSON.

aslakhellesoy:
With this change, `bytes` values have to be encoded as `Encoding::ASCII_8BIT`, which in Ruby means "binary" encoding. Any other encoding will cause the value to be Base64 decoded.\

This is because the JSON representation of `bytes` is [always Base64](https://developers.google.com/protocol-buffers/docs/proto3#json). When deserialising JSON we don't have the protobuf field info, so the decoding is (unfortunately) happening inside `Protobuf::Field::BytesField`, based on whether the value is a "binary" string or a "base64" string.

lib/protobuf/message.rb
```diff
@@ -21,6 +21,22 @@ def self.to_json
...
+      fields = normalize_json(JSON.parse(json))
+      new(fields)
+    end
+
+    def self.normalize_json(ob)
+      case ob
+      when Array
+        ob.map { |value| normalize_json(value) }
+      when Hash
+        Hash[*ob.flat_map { |key, value| [key.underscore, normalize_json(value)] }]
```
film42: The underscore thing seems like we'd want to make that optional?

lib/protobuf/field/bytes_field.rb
```diff
@@ -48,7 +48,18 @@ def wire_type
 
       def coerce!(value)
         case value
-        when String, Symbol
+        when String
+          if value.encoding == Encoding::ASCII_8BIT
```
film42: Would it make sense to make this configurable at the `from_json` method? You can specify if your uses Base64 encoded binary blobs? Maybe true by default but allow the caller to turn it off? That way we don't have to guess.

jackorp:
Ping... Hi, what's the status on this? I am [packaging](https://github.com/fedora-distgit/rubygem-protobuf) this library for Fedora as a part of an effort to update cucumber and it would be nice if this PR and [PR#415](https://github.com/ruby-protobuf/protobuf/pull/415) were merged.

PR #412, opened 2020-01-07
Remove all EOL except 2.3; Add 2.6/2.7 and latest jruby

film42:
Ruby 1.9.3 was failing on the current version of json, which is strange. Most of the versions in the rvm matrix have been EOL'd for a while now. This updates the matrix.

cc @abrandoned @liveh2o


mvz:
Any news on this?

PR #422, opened 2021-06-27
Switch to circle-ci

film42:
This switches our CI to circle-ci. I also went and moved a few Rakefile changes and incorrect proto files changes from #421 to this PR. This was necessary to make CI pass. 

One other note: I did remove a few ruby version from the set. I think 2.4 is the minimum version circle-ci has support for, and honestly, I think that's just fine for now.

@quixoten @abrandoned @mmmries @brianstien @liveh2o


spec/support/protos/map-test.pb.rb
```diff
@@ -12,8 +12,8 @@ module Foo
   # Enum Classes
   #
   class Frobnitz < ::Protobuf::Enum
-    define :FROB, 1
-    define :NITZ, 2
+    define :FROB, 0
```
film42, line 15: The first enum must be zero according to protoc. The map-test.proto and map-test.bin have been updated to make the existing tests pass.

ETetzlaff:
Nice!  LGTM

Issue #427, opened 2022-08-24
Require generated code failed with `const_set': wrong constant name v0 (NameError)


tisonkun:
```
/Users/chenzili/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/protobuf-3.10.6/lib/protobuf/enum.rb:60:in `const_set': wrong constant name v0 (NameError)

      const_set(name, enum)
      ^^^^^^^^^
        from /Users/chenzili/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/protobuf-3.10.6/lib/protobuf/enum.rb:60:in `define'
        from /Users/chenzili/Brittani/pulsar-client-ruby/lib/pulsar/proto/PulsarApi.pb.rb:66:in `<class:ProtocolVersion>'
        from /Users/chenzili/Brittani/pulsar-client-ruby/lib/pulsar/proto/PulsarApi.pb.rb:65:in `<module:Proto>'
        from /Users/chenzili/Brittani/pulsar-client-ruby/lib/pulsar/proto/PulsarApi.pb.rb:9:in `<module:Pulsar>'
        from /Users/chenzili/Brittani/pulsar-client-ruby/lib/pulsar/proto/PulsarApi.pb.rb:8:in `<top (required)>'
        from /Users/chenzili/Brittani/pulsar-client-ruby/bin/console:6:in `require'
        from /Users/chenzili/Brittani/pulsar-client-ruby/bin/console:6:in `<main>'
```

To reproduce:

1. Check out https://github.com/tisonkun/pulsar-client-ruby/tree/prototype
2. Run `bin/setup`
3. Run `bin/console`

tisonkun:
Fixed in https://github.com/tisonkun/pulsar-client-ruby/commit/f829109dd43b074573329779f09852d669f3b445 with `ENV['PB_UPCASE_ENUMS'] = 'true'`.

However, I may want to capitalize instead of upcase.

film42:
Hey! I’m glad you were able to get this issue resolved.

tisonkun:
@film42 what do you think about the capitalize alternative? I have some original enum values like `ServiceNotReady` and I'd prefer to retain them. With `PB_UPCASE_ENUMS` set, it's transformed to `SERVICENOTREADY`. Or it's the convention for Ruby code?

tisonkun:
Here is a quick patch you can review https://github.com/ruby-protobuf/protobuf/pull/428.

[...]

PR #428, opened 2022-08-24
Allow captialize enums name

tisonkun:
Refer to #427.


film42:
Released as `v3.10.7`! Thanks for the contribution.

PR #430, opened 2023-02-28
add healthcheck server

skunkworker:


lib/protobuf/cli.rb
```diff
@@ -236,6 +238,31 @@ def shutdown_server
         ::Protobuf::Rpc::ServiceDirectory.instance.stop
       end
 
+      def start_healthcheck_server
```
skunkworker, line 241: I can move this into it's own class if necessary

film42:
I'm not sure anyone is actually using the service directory anymore, so this might not be the best place for a healthcheck. You should instead use the after server start hook to signal a healthy server to your init system. Otherwise, you should use an initializer to provide app specific insights. For example, MX is using protobuf-nats for transport, and this health check wouldn't guard against a failed NATS cluster, etc.

